PR #54, opened 2026-05-14
ci: Modify the description parameter

xuedinge233:
Modified the description parameter to PR_ID


hipudding (dismissed):
This PR is only used for solve long description blocking other A5 pipeline issue. Skip all logic test because no code change in this PR.

hipudding (approved):

PR #55, opened 2026-05-15
Ci/introduce uv to accelerate pip install for ascend machine

KadenZhang3321:
Replace pip install with uv pip install in CI workflows to speed up dependency resolution.

## Changes

- `integration-tests-ascend.yml` — pip install dist/*.whl → uv pip install

## Improve effect
from 7-9 minutes to 1m30s about 530%

image
 and 
image

 to 

image
and 

image

### DIFF
--- a/.github/workflows/integration-tests-ascend.yml
+++ b/.github/workflows/integration-tests-ascend.yml
@@ -24,6 +24,13 @@ jobs:
     env:
       PROTON_SKIP_PC_SAMPLING_TEST: 1
       PYTHON: "python${{ matrix.python_version }}"
+      UV_INDEX_URL: "http://cache-service.nginx-pypi-cache.svc.cluster.local/pypi/simple"
+      UV_EXTRA_INDEX_URL: "https://repo.huaweicloud.com/ascend/repos/pypi"
+      UV_INDEX_STRATEGY: "unsafe-best-match"
+      UV_INSECURE_HOST: "cache-service.nginx-pypi-cache.svc.cluster.local"
+      UV_HTTP_TIMEOUT: 120
+      UV_NO_CACHE: 1
+      UV_SYSTEM_PYTHON: 1
 
     steps:
       - name: Checkout
@@ -101,8 +108,8 @@ jobs:
           if [ "$NUM_PROCS" -lt 1 ]; then NUM_PROCS=1; fi
           MAX_JOBS="$NUM_PROCS" \
           ${PYTHON} setup.py bdist_wheel
-          ${PYTHON} -m pip install dist/*.whl \
-            -i https://repo.huaweicloud.com/repository/pypi/simple/
+          ${PYTHON} -m pip install uv
+          uv pip install dist/*.whl --no-deps
 
       - name: CCache stats
         run: ccache --print-stats
--- a/.github/workflows/integration-tests-ascend.yml
+++ b/.github/workflows/integration-tests-ascend.yml
@@ -26,7 +26,6 @@ jobs:
       PYTHON: "python${{ matrix.python_version }}"
       UV_INDEX_URL: "http://cache-service.nginx-pypi-cache.svc.cluster.local/pypi/simple"
       UV_EXTRA_INDEX_URL: "https://repo.huaweicloud.com/ascend/repos/pypi"
-      UV_INDEX_STRATEGY: "unsafe-best-match"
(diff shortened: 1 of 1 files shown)


hongziqi (dismissed):

KanuaK (dismissed):

.github/workflows/integration-tests-ascend.yml
```diff
@@ -101,8 +108,8 @@ jobs:
           if [ "$NUM_PROCS" -lt 1 ]; then NUM_PROCS=1; fi
           MAX_JOBS="$NUM_PROCS" \
           ${PYTHON} setup.py bdist_wheel
-          ${PYTHON} -m pip install dist/*.whl \
-            -i https://repo.huaweicloud.com/repository/pypi/simple/
+          ${PYTHON} -m pip install uv
+          uv pip install dist/*.whl --no-deps
```
KanuaK: The image tag already includes the Python version, but the workflow still builds and tests with `${PYTHON}`  explicitly. Should we also bind uv to the same interpreter?

     
     uv pip install --python "${PYTHON}" dist/*.whl --no-deps

This keeps the behavior closer to the previous ${PYTHON} -m pip install ... command and avoids relying on uv's system  Python discovery.

.github/workflows/integration-tests-ascend.yml
```diff
@@ -101,8 +108,8 @@ jobs:
           if [ "$NUM_PROCS" -lt 1 ]; then NUM_PROCS=1; fi
           MAX_JOBS="$NUM_PROCS" \
           ${PYTHON} setup.py bdist_wheel
-          ${PYTHON} -m pip install dist/*.whl \
-            -i https://repo.huaweicloud.com/repository/pypi/simple/
+          ${PYTHON} -m pip install uv
```
KanuaK, line 110: uv itself is installed without a pinned version during the workflow. Would it be better to preinstall a pinned uv version in the Docker image, or at least pin it here, to make CI behavior more reproducible?

.github/workflows/integration-tests-ascend.yml
```diff
@@ -24,6 +24,13 @@ jobs:
     env:
       PROTON_SKIP_PC_SAMPLING_TEST: 1
       PYTHON: "python${{ matrix.python_version }}"
+      UV_INDEX_URL: "http://cache-service.nginx-pypi-cache.svc.cluster.local/pypi/simple"
+      UV_EXTRA_INDEX_URL: "https://repo.huaweicloud.com/ascend/repos/pypi"
+      UV_INDEX_STRATEGY: "unsafe-best-match"
```
KanuaK: Why is UV_INDEX_STRATEGY=unsafe-best-match needed here? Since the actual wheel install uses --no-deps, it seems this may not be necessary for this step.

.github/workflows/integration-tests-ascend.yml
KanuaK: With --no-deps, dependency installation is now delegated to the CI image instead of the wheel metadata. Is the expectation that all runtime/test dependencies from requirements.txt, requirements_dev.txt, and setup.py are already preinstalled in the image? If those dependency lists change later, will the integration-test image be rebuilt and picked up quickly enough?

KadenZhang3321 (commented):
resolveed

.github/workflows/integration-tests-ascend.yml
KadenZhang3321: All runtime and test dependencies (CANN, torch-npu, etc.) are already preinstalled in the CI image by the runner-preparation pipeline. We intentionally use --no-deps to prevent uv from pulling upstream packages and overwriting NPU-specific builds. When dependency lists change, the CI image is rebuilt accordingly.

KadenZhang3321: resolved

.github/workflows/integration-tests-ascend.yml
KadenZhang3321: resolved

.github/workflows/integration-tests-ascend.yml
KadenZhang3321: Currently, uv is only used to install a local wheel (--no-deps), making it very lightweight and having minimal impact from version differences. Pre-installing uv into an image would require additional image maintenance; for now, I guess we can keep `pip install uv` version-unlocked and try to handle this during future image upgrades.

KanuaK (approved):

PR #76, opened 2026-05-17
[release/3.2.2][Autotune] Recover cv dot rhs tuning param inference

zhanwei33:
# `fix: recover cv dot rhs tuning param inference`

## Background

The CV autotune parameter parser relies on dot-site shape inference to recover tunable parameters such as `BLOCK_SIZE_M`, `BLOCK_SIZE_N`, and `BLOCK_SIZE_K`.

For `tl.dot(lhs, rhs)` sites, the parser needs to identify the underlying tensor name and its role on both sides of the dot expression. However, the previous implementation handled only relatively simple RHS expressions. When the RHS tensor was wrapped inside inline arithmetic or conditional expressions, the parser could fail to recover the original tensor role.

As a result, CV kernels using expressions such as:

```python
tl.dot(a, b.to(tl.int8) - tensor_full, out_dtype=tl.int32)
```

could lose the RHS role information, which then caused the `N` dimension tuning parameter inference to break. In practice, this could prevent `BLOCK_SIZE_N` from being preserved in the parsed tunable parameter set.

## Root Cause

The issue came from two related limitations in shape inference:

- Tensor name extraction did not fully support wrapped expressions such as `UnaryOp`, `BinOp`, and `IfExp`.
- Dot-site role resolution did not preserve the tensor role when the RHS expression mixed a real tensor operand with auxiliary scalar or helper expressions.

Because of that, inline RHS expressions could be treated as unresolved, even when the real tensor source was still unambiguous.

## What Changed

This change improves dot-site tensor resolution in `shape_inference.py`:

- Added `_merge_tensor_name_candidates()` to safely merge tensor-name candidates from composite expressions.
- Extended `_extract_tensor_name_from_expr()` to support:
  - `UnaryOp`
  - `BinOp`
  - `IfExp`
- Added `_extract_tensor_name_and_role()` in `_DotSiteCollector` so composite expressions can preserve tensor-role information instead of resolving names only.
- Updated dot-site resolution to select the operand with a valid tensor role when the other side is only an auxiliary expression.
- Kept transpose handling intact by applying role swapping after the base tensor expression is resolved.

With this change, inline RHS expressions such as `b.to(...) - tensor_full` can still resolve back to tensor `b`, allowing the parser to correctly recover the `N`-side tuning parameter.

## Behavior Change

After this fix, CV parameter parsing correctly preserves RHS-derived tuning parameters for inline dot expressions.

For example, in a pattern like:

```python
accumulator += tl.dot(a, b.to(tl.int8) - tensor_full, out_dtype=tl.int32)
```

the parser now continues to infer the RHS `N` dimension and keeps `BLOCK_SIZE_N` in the tunable parameter set.

## Main Files Changed

- `third_party/ascend/backend/runtime/dsl_analysis/shape_inference.py`
- `third_party/ascend/unittest/autotune_ut/test_autotune_param_valid.py`

### DIFF
--- a/third_party/ascend/backend/runtime/dsl_analysis/shape_inference.py
+++ b/third_party/ascend/backend/runtime/dsl_analysis/shape_inference.py
@@ -49,6 +49,17 @@ def _is_tl_transpose_call(node: ast.AST) -> bool:
     )
 
 
+def _merge_tensor_name_candidates(*candidates: Optional[str]) -> Optional[str]:
+    names = []
+    for candidate in candidates:
+        if candidate is None or candidate in names:
+            continue
+        names.append(candidate)
+    if len(names) == 1:
+        return names[0]
+    return None
+
+
 def _extract_tensor_name_from_expr(expr: ast.AST) -> Optional[str]:
     if isinstance(expr, ast.Name):
         return expr.id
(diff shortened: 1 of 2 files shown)


hipudding:
Suggest to change PR title to "cherry-pick: xxxxxx from main" or something better to understand this is a cherry-pick from main branch.

HEX1A0A (approved):

PR #78, opened 2026-05-17
[release/3.2.2][Autotune] Normalize duplicate reduction axis

zhanwei33:
# Fix low-dim parsing failure caused by duplicated reduction-axis prefixes in the Ascend autotuner

## Background

`Q2TritonKernel/tests/mamba/test_angle_dt_kernel.py` failed during the Ascend autotune flow for `angle_dt_bwd_kernel[grid](...)` with the following error:

```text
KeyError: "low dim axis 'rry' not found in known axes: dict_keys(['rx', 'ry'])"
```

The failure happened during autotune axis parsing and tiling-config generation for the backward kernel, rather than during numerical computation.

## Root Cause

The co-debug logs and a local backward-only reproduction showed that the issue came from inconsistent reduction-axis normalization in `autotuner.py`:

- `_get_parser_axis_arg_names()` can already produce reduction-axis names with an `r` prefix, such as `rx` and `ry`
- `_autoparse_reduction_axes()` still unconditionally applies `f"r{axis}"` to the parser output
- As a result, reduction axes that already have the prefix are prefixed again, producing `rrx` / `rry`
- `_autoparse_low_dim_axes()` then continues parsing on the polluted key space, which further pollutes low-dim axes into values such as `rry`
- Finally, `KernelMeta._validate_axis()` receives `low_dim_axes=['rry']` while the known axes are only `rx` / `ry`, which triggers the exception

`tile_generator.py` is not the source of the bug. It only correctly rejects the invalid input propagated from upstream.

## Solution

This fix adopts a unified reduction-axis canonicalization helper in `autotuner.py` and consolidates the reduction-axis naming rules there:

- Add a shared helper to canonicalize reduction-axis names
- Canonicalize only when the name can be reduced to a valid base axis `x/y/z/w/v/t`, instead of using a broad `lstrip("r")`
- Use the following canonicalization rules consistently:
  - `x -> rx`
  - `rx -> rx`
  - `rrx -> rx`
  - `y -> ry`
  - `ry -> ry`
  - `rry -> ry`
- Reuse the same canonicalization logic in both the `vv` path and the legacy parser path

At the same time, `_autoparse_reduction_axes()` is reordered to:

1. Iterate over the raw parser output
2. Extract the base axis
3. Call `_promote_axis_arg_name_to_reduction()` first to complete state migration
4. Generate canonical reduction axes through the shared helper
5. Write the canonical result back to `self.reduction_axes`

This avoids the problem where normalizing first and promoting later could skip state migration for `axis_arg_names` / `vector_axes`.

## Why Not Change `tile_generator.py`

This fix does not add compatibility for `rrx` / `rry` in `tile_generator.py` for the following reasons:

- The downstream exception is a correct safeguard against invalid input
- The real bug is upstream in `autotuner`, where the invalid values are generated
- Relaxing validation downstream would hide the underlying reduction-axis / low-dim naming pollution issue

Therefore, this fix stops the invalid input from being generated in `autotuner.py`, without expanding the scope into a parser protocol redesign or adding fallback behavior in `tile_generator.py`.

## Main Changes

- `third_party/ascend/backend/runtime/autotuner.py`
  - Add a reduction-axis canonicalization helper
  - Reuse the shared helper in the `vv` path
  - Update `_autoparse_reduction_axes()` to promote first and normalize second
- `third_party/ascend/unittest/autotune_ut/test_autotune_param_valid.py`
  - Add minimal coverage for reduction-axis canonicalization
  - Cover the case where parser output already includes the `r` prefix and must not produce `rrx` / `rry`


hipudding (changes requested):
Please describe the specific changes made in the PR in English.

HEX1A0A (approved):

zhanwei33:
Superseded by #146 (https://github.com/triton-lang/triton-ascend/pull/146). Closing this rerolled branch PR in favor of the clean single-commit replacement.

PR #93, opened 2026-05-18
fix(ssbuffer): bugfix Subview results should leads to source.

Skyminers:
# New contributor declaration

  - [x] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests

This PR resolves the following issue: 
When a compute block creates a subview of a memref internally and subsequently manipulates the memory through that subview, the system incorrectly interprets this as an internal memory operation, causing the memory modifications to be lost during external processing.

### DIFF
--- a/third_party/ascend/lib/DynamicCVPipeline/Common/MemoryEffectsTracker.cpp
+++ b/third_party/ascend/lib/DynamicCVPipeline/Common/MemoryEffectsTracker.cpp
@@ -73,6 +73,25 @@ bool isDefinedInside(Value v, Operation *op)
     return op->isProperAncestor(defOp);
 }
 
+Value getViewSource(Value val)
+{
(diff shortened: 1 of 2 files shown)


third_party/ascend/lib/DynamicCVPipeline/Common/MemoryEffectsTracker.cpp
```diff
@@ -73,6 +73,25 @@ bool isDefinedInside(Value v, Operation *op)
     return op->isProperAncestor(defOp);
 }
 
+Value getViewSource(Value val)
```
hongziqi, line 76: This PR is a bug fix. Please start the title with fix: instead of feat:

hongziqi (approved):

PR #106, opened 2026-05-19
fix(ssbufer): Preserve control-flow attributes during canonicalization

Dy045:
Attach a GreedyRewriteConfig listener to transfer control-flow attributes (e.g. hivm.unlikely_condition, ssbuffer.block_id) from original SCF ops to their canonicalized replacements; add a regression test verifying attribute preservation when adjacent zero-result scf.if ops are merged.


Dy045:
yml

Dy045:
re-run ascend950-wheel-build

Dy045:
ascend950-wheel-build

HEX1A0A (approved):

PR #123, opened 2026-05-19
feat(simt): resolve simt_stack_limit with env var + default; always forward

Ritsuka314:
## Summary
Restructure the SIMT per-thread stack-limit configuration so the policy (env var, default `1152`) lives **entirely in triton-ascend**, with `bishengir-compile` / hivmc as straight enforcers of whatever they're told.

## Resolution
`utils._get_simt_stack_limit(per_kernel)`:

    per-kernel `simt_stack_limit`  >  TRITON_SIMT_STACK_LIMIT env  >  1152

- `1152` = CANN runtime's per-thread stack size (new `DEFAULT_SIMT_STACK_LIMIT` constant).
- The env var accepts decimal or hex (`0x...`).
- Negative disables the downstream check; **`0` is a valid (strict) limit, not a disable sentinel**.

## Forwarding
`ttir_to_npubin` now **always** forwards `--simt-stack-limit=<resolved>` to `bishengir-compile` (replaces the prior `if opt.simt_stack_limit:` guard, which also dropped `=0`).

## Docs
Adds `TRITON_SIMT_STACK_LIMIT` and the per-kernel `simt_stack_limit` compiler option to `environment_variable_and_compiler_options_reference.md` (EN + ZH).

## Tests
New unit tests (`test_simt_stack_limit.py`) cover the resolver: default, env var (decimal, hex, `0`, negative, whitespace), per-kernel-wins-over-env precedence, invalid env raises, and invalid env ignored when per-kernel is set.

## Note
**Supersedes the prior `is not None` single-line fix on this PR.** The new design always forwards a value, so the `=0` correctness this PR originally ensured is preserved (0 still flows through unconditionally).

Companion work in NPUIR / HIVMC removes the duplicate default+env handling there (per-layer separation: TA owns policy, NPUIR + hivmc enforce only).


Ritsuka314:
Closing — superseded by a design decision: the SIMT stack-limit policy is being moved entirely to the Triton-Ascend layer.

- **TA** reads `simt_stack_size` from torch_npu's `acl_default.json` and forwards it to the compiler as `--simt-stack-limit`.
- The TA-side `simt_stack_limit` autotune/option knob is **hidden from users**; the value comes solely from `acl_default.json`. When unset there, the default is `1152 bytes/thread`.
- The compiler-side check stays; the end-to-end policy (defaults, sourcing) lives in TA going forward.

A new TA-side patch reading from `acl_default.json` will land separately. The approach in this PR (env var + default; always forward) is no longer the target design.

PR #135, opened 2026-05-20
[Proton](fix)  CMakeLists.txt Cuda, Amd stubs for Ascend compilation

karacupa21:
Make all GPU backends fully optional with pure-CMake stub mechanism

Previously Proton had hard dependencies on NVIDIA CUPTI and AMD ROCTRACER
headers and libraries. This made it impossible to build a clean version
on Ascend (Huawei) or in any CI/CD environment without the full vendor
toolkits installed.

We now treat CUPTI, ROCTRACER and MSPTI as completely optional features
(default OFF). When a backend is disabled:

All vendor-specific .cpp files are excluded from the build
Minimal stub headers are generated in the build directory
include_directories(BEFORE) guarantees that the stubs take precedence
Dummy implementation files provide the exact symbols required by core
code (cuda::getDevice, hip::getDevice, RoctracerProfiler ctor/dtor, etc.)
As a result we can now build a fully functional Proton library on pure
Ascend hardware (or any other environment) without NVIDIA/AMD toolkits,
while still supporting the original backends when the corresponding
option is turned ON.

This gives us a clean CI/CD matrix (cpu-only, cuda, rocm, ascend) and
eliminates the last hard vendor dependencies from the build system.

No changes were made to any .cpp or .h source files — everything is
handled purely in CMake.

https://github.com/triton-lang/triton-ascend/issues/354

# New contributor declaration


karacupa21:
Make all GPU backends fully optional with pure-CMake stub mechanism

Previously Proton had hard dependencies on NVIDIA CUPTI and AMD ROCTRACER
headers and libraries. This made it impossible to build a clean version
on Ascend (Huawei) or in any CI/CD environment without the full vendor
toolkits installed.

We now treat CUPTI, ROCTRACER and MSPTI as completely optional features
(default OFF). When a backend is disabled:

All vendor-specific .cpp files are excluded from the build
Minimal stub headers are generated in the build directory
include_directories(BEFORE) guarantees that the stubs take precedence
Dummy implementation files provide the exact symbols required by core
code (cuda::getDevice, hip::getDevice, RoctracerProfiler ctor/dtor, etc.)
As a result we can now build a fully functional Proton library on pure
Ascend hardware (or any other environment) without NVIDIA/AMD toolkits,
while still supporting the original backends when the corresponding
option is turned ON.

This gives us a clean CI/CD matrix (cpu-only, cuda, rocm, ascend) and
eliminates the last hard vendor dependencies from the build system.

No changes were made to any .cpp or .h source files — everything is
handled purely in CMake.

third_party/proton/CMakeLists.txt
```diff
@@ -1,3 +1,230 @@
+# =============================================================================
```
hongziqi, line 1: Thanks for your contribution. Since files under `third_party/proton` belong to upstream community assets, we do not recommend modifying them. Such changes will introduce maintenance overhead during future version upgrades.

xqnmqx: Otherwise in case of AscendNPU development you will have a deal with CUDA dependencies. Why we must download them?

xqnmqx: And AMD dependencies too.

PR #151, opened 2026-05-20
[debug](feat) Add source-level stepping in msdebug

MrBambastik:
feat(ascend): Add source-level stepping in msdebug

Make Triton kernel source lines breakable in msdebug on Ascend NPU and
make `next` stop once per source line. Opt-in via
LLVM_EXTRACT_DI_LOCAL_VARIABLES; with the flag unset, codegen is
bit-identical and the binary is left untouched.

Problem
-------
Reaching a kernel binary optimizes many Triton ops away: program-id and
num-programs become argument reads, addptr lowers to no machine code,
tl.full splats are constant-folded, and tl.arange is absorbed into a
reinterpret_cast. Each vanished op takes its PC->source-line mapping
with it, so msdebug has no DWARF entry to break on. The lines that do
survive are then emitted with duplicate, phantom, and scattered rows, so
a single `next` stops several times on one line or steps into inlined
libdevice code.

Approach
--------
Preserve the lines, then clean them, in three stages all gated on the
env var:

1. Preserve (TritonToLinalg). Before an affected op is replaced, emit an
   llvm.inline_asm "nop" tagged with its source location so the line
   keeps a real PC and DWARF entry (DebugUtils.h::insertDebugNop;
   converters for pid/num_programs/addptr/splat/make_range, plus the
   BlockPtrAnalysis path where make_range is absorbed). SplatOp::fold is
   gated so a scalar splat survives to its converter, and FusedLocs are
   unwrapped to a concrete line.

2. Clean locations (MLIR). CanonicalizeDebugLocationsPass normalizes
   fused/callsite locations and collapses duplicates;
   DeduplicateDebugNopsPass drops redundant consecutive NOP anchors.

3. Clean the line table (DWARF). debug_line_rewriter.py post-processes
   .debug_line so `next` stops once per user line, demoting duplicate,
   foreign-file, line-0, and out-of-range is_stmt rows (the last catches
   phantom lines bishengir pulls in from inlined libdevice) while
   preserving loop back-edges. The edit is length-preserving
   (DW_LNS_copy -> DW_LNS_set_basic_block) so .rela.debug_line stays
   valid; it is atomic and self-verifying, returning the binary
   unchanged on any error.

compiler.py enables bishengir's debug-info/ms-debug emission and wraps
the npubin stage with the rewriter, both under the same flag and no-ops
otherwise.

Tests
-----
Lit: debug_nop.mlir (NOP insertion, both modes) and dedup_nop.mlir
(redundant-NOP removal). Unit: test_debug_line_rewriter.py covers the
rewriter end-to-end (31 cases, hardware-free).

# New contributor declaration

  - [x] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests


xqnmqx:
Could you write tests in your code?

xqnmqx:
@MrBambastik , please use tools::getBoolEnv("TRITON_DEBUG") instead of direct usage getEnv method.

PR #159, opened 2026-05-21
[CI] Rename the display of the Ascend950 Pipeline Tests

xuedinge233:
Summary
------------
Since the `Ascend pipeline tests` task is triggered by `workflow_run`, it does not automatically obtain the pr name. Here, the run-name method is used to synchronize the task names in the action list to the pr names. At the same time, identifiers have been added to the task names related to `workflow_dispatch` tasks and `cleanup` tasks


xuedinge233:
This is the test result of fork for three different scenarios:
image

hipudding (approved):
Skip 950 test check since only change github workflow

PR #167, opened 2026-05-21
fix(cdiv): wrap result in constexpr for Number inputs

tarvlad:
Prevent type errors in cases like `range(tl.cdiv(X, Y) + 1, ...`


WuTYSFG (approved):
Work around the bug in our custom cdiv implementation (originally added to support Triton Ascend specs). This PR fixes the issue as a temporary measure. We plan to remove this custom implementation and align with the upstream community cdiv in the future.

PR #197, opened 2026-05-23
feat(mask): solve the tile overlapping issue when grid launch parllel

liufengsheng-beep:
Add traceUserToTargetOp  func to trace the use-def chains of tt.get_program_id. It verifies that program id are used in a constant multiplication + addition pattern that matches the range from tt.make_range, ensuring non-overlapping memory access between blocks. 
only the type like this: program_id * BLOCK_SIZE + BLOCK_SIZE can be regarded as tile non-overlap, otherwise, it will be regarded as tile overlap, and set the syncblocklock to be true to run test cases serially in discrete mask template.

# New contributor declaration

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests

### DIFF
--- a/third_party/ascend/lib/DiscreteMaskAccessConversion/DiscreteMaskAccessConversionPass.cpp
+++ b/third_party/ascend/lib/DiscreteMaskAccessConversion/DiscreteMaskAccessConversionPass.cpp
@@ -57,6 +57,126 @@ static bool enableSyncBlockLockFlag = true;
 static constexpr const char *routeDiscreteMaskToSimtAttrName =
     "route_discrete_mask_to_simt";
 
+static bool traceUserToTargetOp(Value val) {
+  llvm::SmallVector<Value, 32> worklist;
+  llvm::SmallPtrSet<Value, 32> visited;
+  worklist.push_back(val);
+
+  while (!worklist.empty()) {
+      Value currVal = worklist.pop_back_val();
+      if (!visited.insert(currVal).second)
+          continue;
+
+      for (Operation* user : currVal.getUsers()) {
+          if (auto mulOp = dyn_cast<arith::MulIOp>(user)) {
+              Value lhs = mulOp.getLhs();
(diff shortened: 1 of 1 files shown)


third_party/ascend/lib/DiscreteMaskAccessConversion/DiscreteMaskAccessConversionPass.cpp
```diff
@@ -57,6 +57,126 @@ static bool enableSyncBlockLockFlag = true;
...
+                    
+                        Value curSrc = otherOperand;
+                        bool hitRange = false;
+                        int depth = 0;
+                        while (curSrc.getDefiningOp() && depth < 5)
+                        {
+                            Operation* defOp = curSrc.getDefiningOp();
+                            if (auto rangeOp = dyn_cast<triton::MakeRangeOp>(defOp))
+                            {
+                                if (rangeOp.getEnd() == blockSize)
```
kaixin1976, line 113: hit RangeOp is not sufficent, should hit the addPtr used by the StoreOp which has discrete mask . That is to say the storeOp is using an independant block, then no need to lock , otherwise , we need to lock .

third_party/ascend/lib/DiscreteMaskAccessConversion/DiscreteMaskAccessConversionPass.cpp
```diff
@@ -348,7 +468,8 @@ DiscreteMaskAccessConversionPass::DiscreteMaskAccessConversionPass(
 void DiscreteMaskAccessConversionPass::runOnOperation() {
   compileOn91095Flag = this->compileOn91095;
   forceSimtTemplateFlag = this->forceSimtTemplate;
-  enableSyncBlockLockFlag = this->enableSyncBlockLock;
```
kaixin1976, line 351: NonOverlap checking should do only for storeOp with discrete mask .

PR #235, opened 2026-05-25
(fix) Export CANN vector ops from triton.language

L4-1024:
## Summary
- Export CANN vector ops `extract_slice`, `get_element`, and `insert_slice` from `triton.language`.
- Add a focused import-level regression test that keeps the `tl.*` aliases aligned with `triton.language.extra.cann.extension`.

## Motivation
Some compatibility code and kernels reference CANN vector ops through the common `triton.language as tl` namespace. The CANN extension already exports these ops, but `triton.language` does not expose them, which makes the `tl.*` path fail.

Evidence from the reported environment:
```text
triton file: /opt/mamba/envs/ascend-infer/lib/python3.13/site-packages/triton/__init__.py
tl file: /opt/mamba/envs/ascend-infer/lib/python3.13/site-packages/triton/language/__init__.py
has tl.extract_slice: False
has al.extract_slice: True
al file: /opt/mamba/envs/ascend-infer/lib/python3.13/site-packages/triton/language/extra/cann/extension/__init__.py
```

The missing `tl.extract_slice` also appears in runtime failure traces when JIT visits an attribute access:
```text
File "/opt/mamba/envs/ascend-infer/lib/python3.13/site-packages/triton/runtime/jit.py", line 144, in visit_Attribute
AttributeError: module 'triton.language' has no attribute 'extract_slice'
Error 模型 Qwen/Qwen3-30B-A3B 验证失败
```

Exporting these existing CANN extension functions from `triton.language` restores the expected `tl.extract_slice` compatibility path without changing the underlying implementation.

## Validation
- `python3 -m py_compile python/triton/language/__init__.py third_party/ascend/unittest/pytest_ut/test_language_exports.py`
- `git diff --check upstream/main...fix/export-cann-vec-ops-github`

## Note
- `PYTHONPATH="$PWD/python:$PWD/third_party/ascend" /opt/mamba/envs/ascend-infer/bin/python -m pytest -q third_party/ascend/unittest/pytest_ut/test_language_exports.py` is blocked in this local checkout because the source tree has not built `triton._C`; CI/build environments with the extension available should run the added test normally.

### DIFF
--- a/python/triton/language/__init__.py
+++ b/python/triton/language/__init__.py
@@ -131,6 +131,11 @@ from .random import (
     randn4x,
     uint_to_uniform_float,
 )
+from .extra.cann.extension import (
+    extract_slice,
+    get_element,
+    insert_slice,
+)
 from . import target_info
 
 __all__ = [
@@ -183,6 +188,7 @@ __all__ = [
     "erf",
     "exp",
     "exp2",
+    "extract_slice",
     "expand_dims",
     "extra",
     "fdiv",
@@ -199,10 +205,12 @@ __all__ = [
     "fma",
     "full",
     "gather",
+    "get_element",
     "histogram",
     "inline_asm_elementwise",
     "interleave",
     "int1",
+    "insert_slice",
     "int16",
     "int32",
     "int64",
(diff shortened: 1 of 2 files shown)


L4-1024:
@HEX1A0A

python/triton/language/__init__.py
```diff
@@ -131,6 +131,11 @@
     randn4x,
     uint_to_uniform_float,
 )
+from .extra.cann.extension import (
```
hongziqi, line 134: It is not recommended to make invasive modifications to community code

PR #247, opened 2026-05-26
fix: extractOp is added to be marked as Datause in UseAnalysis

WuTYSFG:
The use analysis lacks an explicit marking for extractOp, which may lead to the erroneous marking of extractOp as MetaUse. Since the result of extractOp is a scalar, pointer analysis will stop when it identifies a scalar. Therefore, if extractOp is involved in pointer calculations, it must be explicitly marked as DataUse. If this is not done, it may be incorrectly deleted by MetaUseEraser, introducing a series of issues related to unrealized_conversion_cast. 

# New contributor declaration

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests


kaixin1976 (approved):
look good to me

PR #257, opened 2026-05-26
[release/3.2.2-dev] test_fused_moe_router_tensorcore_kernel passed

pavel-pikunov:
This change works around the accuracy issue in fused_moe_router_tensorcore_kernel by folding a redundant one-hot value selection in TTAdapter IR.

The generated IR reconstructed the selected top2 value through:

cmpi -> uitofp -> mulf -> reduce_sum

although max_with_index had already produced both the selected value and index. On Ascend this path can produce an incorrect top2 value, leading to a wrong top2 weight.

The workaround reuses the value result of max_with_index directly instead of recomputing it through the one-hot mask. This fixes the failing test, but does not fully address the underlying lower-level issue in the mask/cast/multiply/reduce path.

# New contributor declaration

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests
  - [x] This PR does not need a test because `FILL THIS IN`.

### DIFF
--- a/third_party/ascend/lib/TritonToLinalg/TritonToLinalgPass.cpp
+++ b/third_party/ascend/lib/TritonToLinalg/TritonToLinalgPass.cpp
@@ -766,6 +766,140 @@ LogicalResult TritonToLinalgPass::processLegalStrideOperations(ModuleOp moduleOp
   return success();
 }
 
+struct FoldOneHotGatherAfterReduceWithIndex
+    : public OpRewritePattern<linalg::ReduceOp> {
+  using OpRewritePattern<linalg::ReduceOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(linalg::ReduceOp op,
+                                PatternRewriter &rewriter) const final {
+    // Match only ordinary reduce-sum, not max_with_index itself.
+    if (op->getAttrOfType<StringAttr>("reduce_mode"))
+      return failure();
+
+    if (op.getNumResults() != 1 || op.getInputs().size() != 1)
+      return failure();
+
+    if (!isAddFReduction(op))
+      return failure();
+
+    auto mulOp = op.getInputs()[0].getDefiningOp<arith::MulFOp>();
+    if (!mulOp)
+      return failure();
+
+    Value logits;
+    Value maskAsFloat;
+
+    if (mulOp.getRhs().getDefiningOp<arith::UIToFPOp>()) {
+      logits = mulOp.getLhs();
+      maskAsFloat = mulOp.getRhs();
+    } else if (mulOp.getLhs().getDefiningOp<arith::UIToFPOp>()) {
+      logits = mulOp.getRhs();
+      maskAsFloat = mulOp.getLhs();
+    } else {
+      return failure();
+    }
+
+    auto castOp = maskAsFloat.getDefiningOp<arith::UIToFPOp>();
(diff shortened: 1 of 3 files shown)


third_party/ascend/lib/TritonToLinalg/TritonToLinalgPass.cpp
```diff
@@ -904,6 +1038,20 @@ void TritonToLinalgPass::runOnOperation() {
     signalPassFailure();
   }
 
+// 7.1 Fold redundant one-hot gather after max_with_index.
```
WuTYSFG: Why is linalg::reduceOp handled at the end, instead of handling triton::reduceOp in CanonicalizationPatterns?

third_party/ascend/lib/TritonToLinalg/TritonToLinalgPass.cpp
```diff
@@ -766,6 +766,140 @@ LogicalResult TritonToLinalgPass::processLegalStrideOperations(ModuleOp moduleOp
   return success();
 }
 
+struct FoldOneHotGatherAfterReduceWithIndex
```
WuTYSFG: OpRewritePattern needs to be placed in xxxConverter.cpp.

WuTYSFG: If it is necessary to convert linalg::reduceOp, there is currently no suitable file to accommodate it. If a new linalgOpConverter.cpp needs to be created, it is necessary to consult our architect for confirmation. @kaixin1976

third_party/ascend/lib/TritonToLinalg/TritonToLinalgPass.cpp
pavel-pikunov: This fix is intentionally a post-conversion workaround on linalg::ReduceOp, where the problematic duplicated pattern is explicit and can be safely folded away.

third_party/ascend/lib/TritonToLinalg/TritonToLinalgPass.cpp
pavel-pikunov: Agreed. The pattern is kept in an existing converter/canonicalizer location (ArgMinMaxConverter.cpp), not in TritonToLinalgPass.cpp.
The pass only performs registration/invocation.

pavel-pikunov: The pattern is not a general Linalg reduce rewrite; it specifically handles the value/index pair produced by ArgMin/ArgMax lowering. For that reason, placing it near the existing ArgMinMax converter seems like the most localized option for now.

pavel-pikunov: Code functionality verified. Target test passed.

pavel-pikunov: Code functionality verified. Target test passed.

third_party/ascend/lib/TritonToLinalg/TritonToLinalgPass.cpp
pavel-pikunov: Code functionality verified. Target test passed.

WuTYSFG (approved):

PR #288, opened 2026-05-28
feat(autoblockify): Enable TRITON_ALL_BLOCKS_PARALLEL by default

zackc6:
feat(autoblockify): Enable TRITON_ALL_BLOCKS_PARALLEL by default


third_party/ascend/unittest/pytest_ut/test_dot_scaled.py
```diff
@@ -38,6 +38,7 @@
 from triton.language.extra import libdevice
 
 
+@pytest.mark.skip(reason="skip on this case")
```
melo882: Why skip all the test case in dot_scaled? Are there any cases can be passed in this case?

zackc6: Thanks for reminder, changed to only  skip one case :)

third_party/ascend/unittest/pytest_ut/test_triton_unified_attention.py
```diff
@@ -384,6 +384,7 @@ def ref_paged_attn(
     return torch.cat(outputs, dim=0)
 
 
+@pytest.mark.skip(reason="skip on this case")
```
WuTYSFG, line 387: Why will we skip this case

WuTYSFG (approved):
The skipped test case has been submitted on dts for tracking

PR #296, opened 2026-05-28
docs: clarify that multibuffer is enabled by default

YujunHuang063:
## Summary

Update the EN and ZH compiler options reference tables (`environment_variable_and_compiler_options_reference.md`) and the architecture / feature overview tables (`architecture_design_and_core_features.md`) to explicitly state that `multibuffer` is enabled by default.

The previous wording either left the default implicit or only mentioned a 910_95 caveat, which could confuse users about the actual default value of this option.

This is a docs-only change.

## Test plan

- [ ] Verify docs render correctly in the documentation site.
- [ ] Confirm no functional code changes (docs-only PR).


kaixin1976 (approved):
lgtm

PR #499, opened 2026-06-09
[CI](chore) Separate build and test

xuedinge233:
### Summary

- Build the TA wheel package into the cpu, and the npu only runs the test process to reduce the resource occupation of the npu
- MLIR FileCheck tests are pure CPU compile-checks, put it into the cpu as well


hipudding:
Close this PR due to no improvement found by separating build and test. Reopen it if need.

PR #517, opened 2026-06-10
[Frontend](fix) Revert fp8 datatype support for tl.full

TecJesh:
### Related Issue
#645 
### Summary

Remove the `get_fp8e4nv` and `get_fp8e5` intrusive modifications from `ir.cc` that enabled fp8 constant value creation.

### Why

These two functions were originally introduced as an intrusive change to support fp8 data types in `tl.full`,
However, the `full` op spec does not support fp8, remove it.

# New contributor declaration

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests

### DIFF
--- a/python/src/ir.cc
+++ b/python/src/ir.cc
@@ -932,16 +932,6 @@ void init_triton_ir(py::module &&m) {
              return self.create<arith::ConstantOp>(
                  self.getBuilder().getF64FloatAttr(v));
            })
-      .def("get_fp8e4nv",
-           [](TritonOpBuilder &self, double v) -> Value {
-             return self.create<arith::ConstantOp>(FloatAttr::get(
-                 Float8E4M3FNType::get((self.getBuilder().getContext())), v));
-           })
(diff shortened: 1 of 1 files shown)


python/src/ir.cc
```diff
@@ -932,16 +932,6 @@ void init_triton_ir(py::module &&m) {
              return self.create<arith::ConstantOp>(
                  self.getBuilder().getF64FloatAttr(v));
            })
-      .def("get_fp8e4nv",
```
hongziqi, line 935: check with WangTao

hongziqi (approved):

python/src/ir.cc
KanuaK: show checked result here @TecJesh

TecJesh: tl.full does not support fp8，revert to the community implementation

KanuaK (approved):

PR #530, opened 2026-06-11
[ci](feat) label restricted file changes

tianli-Backkom:
## Summary
- Add a dedicated `pull_request_target` workflow that labels PRs changing files outside repository-construction paths.
- Allowed directories: `.github/`, `docker/`, `docs/`, `third_party/`.
- Allowed root files: `.readthedocs.yaml`, `CONTRIBUTING.md`, `CONTRIBUTING_zh.md`, `README.md`, `README_zh.md`, `RELEASE.md`, `SECURITYNOTE.md`, `SECURITYNOTE_zh.md`, `requirements.txt`, `requirements_dev.txt`, `safe_compile.cmake`.
- Use the GitHub PR files API without checking out or executing fork code.

## Behavior
- If a PR changes files outside the allowed paths, the workflow applies the `restricted-files` label and exits successfully.
- If all changed files are within the allowed paths, the workflow removes `restricted-files` when present and exits successfully.
- The workflow is split into scan/apply/remove steps so each task has a single responsibility.

## Review feedback addressed
- Use the default `pull_request_target` activity types instead of listing unnecessary types.
- Move `SECURITYNOTE.md`, `SECURITYNOTE_zh.md`, and `safe_compile.cmake` to the allowed-file list.
- Remove duplicate allowed path printing from the success branch.
- Remove the separate `restricted_files` list; the rule is now based only on allowed directories and files.
- Split the workflow by function: scan changed files, apply label, remove label.
- Grant `pull-requests: write` so the labeler can update PR labels.

## Validation
- Parsed the workflow YAML locally with PyYAML.
- Ran `bash -n` against each workflow shell step.
- Ran `pre-commit run check-yaml --files .github/workflows/protected-files-check.yml` successfully.
- Ran `git diff --check` successfully.
- Verified in the personal fork:
  - PR #3 changes `.pre-commit-config.yaml` and receives `restricted-files`.
  - PR #4 changes only `docs/` and does not receive `restricted-files`.


.github/workflows/protected-files-check.yml
```diff
@@ -0,0 +1,142 @@
+name: Restricted Files Labeler
+
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened, ready_for_review, edited]
```
hipudding: Do we need all these types?

.github/workflows/protected-files-check.yml
```diff
@@ -0,0 +1,142 @@
...
+        shell: bash
+        env:
+          GH_TOKEN: ${{ github.token }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          REPOSITORY: ${{ github.repository }}
+          RESTRICTED_FILES_LABEL: restricted-files
+        run: |
+          set -euo pipefail
+
+          restricted_files=(
```
hipudding: I think these file are allowed to modify.

.github/workflows/protected-files-check.yml
```diff
@@ -0,0 +1,142 @@
...
+              "repos/${REPOSITORY}/issues/${PR_NUMBER}/labels" \
+              -f "labels[]=${RESTRICTED_FILES_LABEL}" \
+              >/dev/null
+
+            echo "Applied label: ${RESTRICTED_FILES_LABEL}"
+            exit 0
+          fi
+
+          echo "All changed files are within the allowed repository-construction paths."
+          echo "Restricted files:"
```
hipudding: Duplicate with line 97-102. I think we don't need print these files for dirs, contributor can read this yaml for detail.

xuedinge233:
The current tasks are all mixed into one step. It can be considered to split them by function.

hipudding (dismissed):

hipudding (dismissed):

hipudding (approved):

KanuaK (approved):

PR #534, opened 2026-06-11
[ssbuffer](fix): restore control-attr transfer and split split matmul pattern adjustments

Dy045:
Replace nested-op exclusion with same-block check for fallback attribute transfer candidates, and add null guard for oldOp block.

### DIFF
--- a/third_party/ascend/lib/DynamicCVPipeline/SplitDataflow/PreserveControlAttrsCanonicalize.cpp
+++ b/third_party/ascend/lib/DynamicCVPipeline/SplitDataflow/PreserveControlAttrsCanonicalize.cpp
@@ -95,16 +95,17 @@ private:
                 return defOp;
         }
 
-        // Fallback: search recentInserts in reverse, skip nested candidates
+        Block *oldBlock = oldOp->getBlock();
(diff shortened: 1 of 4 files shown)


Skyminers (approved):

KanuaK (approved):

third_party/ascend/lib/DynamicCVPipeline/SplitDataflow/PreserveControlAttrsCanonicalize.cpp
```diff
@@ -95,16 +95,17 @@ class PreserveControlAttrsListener : public RewriterBase::Listener {
                 return defOp;
         }
 
-        // Fallback: search recentInserts in reverse, skip nested candidates
+        Block *oldBlock = oldOp->getBlock();
```
hongziqi, line 98: It is recommended to supplement the PR description.

Dy045: 已修改

hongziqi (approved):

PR #582, opened 2026-06-15
[Frontend](fix) Remove original exception info retention

TecJesh:
### Related Issue
#645 

### Summary

Revert `code_generator.py` to preserve original exception info via `repr(e)`, and set `TRITON_FRONT_END_DEBUGGING=1` in `test_isfinited_invalid_dtype` to maintain correct error propagation.

### Background

The frontend error wrapping in `code_generator.py` was previously modified to drop or simplify exception info (`repr(e)` → `None` / `str(e)`). This caused the `CompileTimeAssertionFailure` raised by `libdevice.isfinited()` with an invalid dtype to lose its original error message, breaking `test_isfinited_invalid_dtype`.

This PR reverts those changes to preserve the full original error, and instead uses the `TRITON_FRONT_END_DEBUGGING=1` environment variable to achieve the same effect: bypassing the `CompilationError` wrapper so the original `CompileTimeAssertionFailure` propagates directly, where `raises_with_match` can capture and verify it.

### User Impact

Users who need to inspect the original exception traceback without the `CompilationError` wrapper can set:

```bash
export TRITON_FRONT_END_DEBUGGING=1
```
### Test Result
image

# New contributor declaration

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests

### DIFF
--- a/third_party/ascend/unittest/pytest_ut/test_isfinited.py
+++ b/third_party/ascend/unittest/pytest_ut/test_isfinited.py
@@ -107,7 +107,8 @@ invalid_types = [
 @pytest.mark.parametrize("sigtype", invalid_types)
 @pytest.mark.parametrize("N", shapes)
 @test_common.raises_with_match(triton.compiler.errors.CompilationError, "Expected dtype fp16/fp32/bf16, but got")
-def test_isfinited_invalid_dtype(sigtype, N):
+def test_isfinited_invalid_dtype(sigtype, N, monkeypatch):
+    monkeypatch.setenv('TRITON_FRONT_END_DEBUGGING', '1')
 
     def torch_func(x0):
         res = torch.isfinite(x0)
--- a/python/triton/compiler/code_generator.py
+++ b/python/triton/compiler/code_generator.py
@@ -1353,7 +1353,7 @@ class CodeGenerator(ast.NodeVisitor):
                 # Wrap the error in the callee with the location of the call.
                 if knobs.compilation.front_end_debugging:
                     raise
-                raise CompilationError(self.jit_fn.src, self.cur_node, repr(e)) from e
+                raise CompilationError(self.jit_fn.src, self.cur_node, None) from e
 
             callee_ret_type = generator.ret_type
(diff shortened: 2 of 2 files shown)


third_party/ascend/unittest/pytest_ut/test_isfinited.py
```diff
@@ -107,7 +107,8 @@ def triton_func(x0, N):
 @pytest.mark.parametrize("sigtype", invalid_types)
 @pytest.mark.parametrize("N", shapes)
 @test_common.raises_with_match(triton.compiler.errors.CompilationError, "Expected dtype fp16/fp32/bf16, but got")
-def test_isfinited_invalid_dtype(sigtype, N):
+def test_isfinited_invalid_dtype(sigtype, N, monkeypatch):
+    monkeypatch.setenv('TRITON_FRONT_END_DEBUGGING', '1')
```
hongziqi, line 111: give result on pr

hongziqi, line 111: If the environment variable is not set, what is the error message?

TecJesh: If the env variable is not set, the error message is `'at 4:10:\ndef triton_kernel(out_ptr0, in_ptr0, N: tl.constexpr):\n    idx = tl.arange(0, N)\n    x0 = tl.load(in_ptr0 + idx)\n    ret = libdevice.isfinited(x0)\n`, which can not expose the original exception message. the error message should be `Expected dtype fp16/fp32/bf16, but got int32`

hongziqi: Does this error message phenomenon align with the GPU?

TecJesh: Verified, the error message phenomenon aligns with the GPU

hongziqi (approved):

KanuaK (approved):

PR #584, opened 2026-06-15
[Language](fix) Revert fast_math to False

TecJesh:
### Related Issue
#645 
### Summary

Revert `dot_scaled`'s `fast_math` default to `False` to align with upstream Triton 3.5, and add explicit `fast_math=True` to existing test cases that were written before this parameter was introduced.

### Background

`fast_math` is a new parameter on `tl.dot_scaled` introduced in Triton 3.5. Its default is `False`, which enables NaN sentinel value checking in micro-scaling (MX) format, the corresponding data position is set to NaN.

Previously, the fast_math parameter changed the default to `True` to avoid breaking existing Ascend test cases. Those tests were written before `fast_math` existed and don't account for NaN checking, which would trigger NaN injection and cause precision mismatches with the reference (`golden_ref` does not handle NaN sentinels). By default, dot_scaled is supposed to check for NaN. To skip NaN checking, one must explicitly specify fast_math=True. Changing the default value of fast_math is not correct.

### User Impact

Existing code that calls `tl.dot_scaled` without explicitly passing `fast_math` will now use `fast_math=False` (the new default). This means:

- NaN will be checked at some positions.
- Users should explicitly pass `fast_math=True` if their use case does not require NaN sentinel checking.

# New contributor declaration

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests


python/triton/language/core.py
```diff
@@ -2031,7 +2031,7 @@ def dot(input, other, acc=None, input_precision=None, allow_tf32=None, max_num_i
 
 
 @builtin
-def dot_scaled(lhs, lhs_scale, lhs_format, rhs, rhs_scale, rhs_format, acc=None, fast_math=True, lhs_k_pack=True,
+def dot_scaled(lhs, lhs_scale, lhs_format, rhs, rhs_scale, rhs_format, acc=None, fast_math=False, lhs_k_pack=True,
```
melo882, line 2034: 为何要对fast_math参数默认值进行修改？

hongziqi (approved):

python/triton/language/core.py
TecJesh: As shown in the PR description.

WuTYSFG (approved):

PR #587, opened 2026-06-15
[annotate-user-casts](feat) annotate casts written by user

tarvlad:
## Background

The backend needs to distinguish casts that were present in the original Triton IR from casts introduced later by compiler transformations. This is needed so backend optimizations can safely reason about compiler-inserted cast chains without accidentally changing user-written cast semantics.

## What changed

This PR introduces a new `annotate-user-casts` pass that marks cast operations already present at this stage of the pipeline with:

```mlir
cast.source = "user"
```


hongziqi (approved):

third_party/ascend/backend/compiler.py
```diff
@@ -194,6 +194,7 @@ def ttir_to_linalg(mod, metadata, opt, *, named_ops=False):
             enable_sync_block_lock
         )
         ascend.passes.ttir.add_triton_to_annotation(pm)
+        ascend.passes.ttir.add_annotate_user_casts(pm)
```
hongziqi, line 197: It is recommended that you clearly describe the background and purpose of the PR for tagging users and the final effect.

tarvlad: done, pls check pr description

PR #608, opened 2026-06-16
[costmodel](fix) optimize costmodel precision

BloodOrange666:
## Motivation

The previous costmodel could significantly under-estimate kernels where important work is hidden behind helper functions called inside loops. This affected configuration ranking during costmodel-based filtering and could cause slower configs to be selected.

By inlining helper functions before estimation and using the loop-weighted roofline summary, the reported estimate better reflects the actual repeated work in the kernel.

## Summary

This PR improves the precision of the Ascend cost model by aligning the in-process estimation pipeline with loop-aware pipeline analysis and Ascend 910B hardware behavior.

The main fix is to avoid under-estimating operations that are placed in Triton helper functions, such as reduction helpers used by softmax. These helpers can be called from inside loops, but without inlining their operations are analyzed at helper-function scope and receive a loop multiplier of `1`. This causes per-iteration work to be under-counted during costmodel prefiltering.

## Changes

- Run MLIR inlining and symbol DCE before the in-process costmodel conversion pipeline.
  - This lets `tt.call` helper bodies be analyzed at the actual call site.
  - Dead helper symbols are removed after inlining to avoid double-counting.

- Make costmodel result extraction prefer module-level scheduled/roofline summaries.
  - `ascend.scheduled_cycles`
  - `ascend.roofline_cycles`
  - `ascend.simple_sum_cycles`
  - fallback to per-op `estimated_cycles * loop_multiplier`

- Update pipeline analysis to publish loop-multiplied roofline cycles as `ascend.scheduled_cycles`.

- Model Ascend 910B vector-side transfer serialization more accurately by accounting for the mutex between vector MTE2 load and MTE3 store paths.

- Add memory-space information to inserted transfer ops so transfer modeling can distinguish source and destination spaces.

### DIFF
--- a/third_party/ascend/backend/runtime/costmodel_runtime.py
+++ b/third_party/ascend/backend/runtime/costmodel_runtime.py
@@ -74,7 +74,7 @@ def make_costmodel_cache_key(ttir: str, extra_args: Optional[List[str]]) -> str:
     if extra_args:
         h.update(" ".join(extra_args).encode("utf-8"))
     h.update(b"|")
-    h.update(b"inproc_costmodel_v1")
+    h.update(b"inproc_costmodel_v2_loop_weighted")
     return h.hexdigest()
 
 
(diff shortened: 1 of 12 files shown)


third_party/ascend/backend/runtime/costmodel_runtime.py
```diff
@@ -74,7 +74,7 @@ def make_costmodel_cache_key(ttir: str, extra_args: Optional[List[str]]) -> str:
     if extra_args:
         h.update(" ".join(extra_args).encode("utf-8"))
     h.update(b"|")
-    h.update(b"inproc_costmodel_v1")
+    h.update(b"inproc_costmodel_v2_loop_weighted")
```
hongziqi, line 77: It is recommended to supplement a detailed PR description, including the background, effects, changes, etc.

BloodOrange666: supplemented

HinPeng (approved):

PR #656, opened 2026-06-18
[ut](fix) fix tl.parallel ut

CHNJZ:
### Background

`test_parallel.py::test_add_bind_false` has been skipped long-term via `@pytest.mark.skip(reason="not supported after the NPUIR is updated in April")` since the NPUIR update, leaving `extension.parallel` without effective UT coverage.

### Root Cause

The skipped `triton_add` kernel **misuses `extension.parallel` to express a serial accumulation**, violating the parallel-loop contract:

- The loop body `ret = ret + x1` forms a loop-carried dependency, making `ret` an `scf.for` iter_arg, which contradicts the semantics of `hivm.parallel_loop` (which asserts that iterations are independent).
- Every iteration operates on the entire `ret` block and hits the same UB address, failing the "disjoint memory footprint across iterations" requirement.
- The induction variable is not used for addressing; the loop merely repeats the accumulation N times.

This pattern only produced correct results incidentally, while the backend executed the parallel loop serially. Once the NPUIR update began enforcing/validating parallel semantics, it was no longer supported. The case was fundamentally invalid — this is not a backend regression.

### Changes

Rewrite the kernel into a **contract-compliant** parallel use case that uses `extension.parallel` as designed (following the CV-fusion `extract_slice` partitioning idiom): use the induction variable `s` to take **disjoint** sub-blocks, perform a single compute instruction in the body, and write back to a **disjoint** output region, using a **signed integer** dtype that is on the validated-support list.

Key changes relative to the old case:

1. **Eliminate the loop-carried dependency**: no longer accumulate into a shared `ret`; instead each iteration independently processes one sub-block.
2. **Guarantee disjoint footprints**: partition both the input sub-block and the output region by `s * SUB_M`.
3. **Converge to ≤1 compute instruction**: the loop body keeps only a single `+ 1`.
4. **Remove `bind_sub_block=True`**: this parameter is read and then discarded in the frontend, produces no IR difference, and is a dead/legacy option — keeping it is only misleading.
5. **Add a parallel-marker assertion**: assert the `hivm.parallel_loop` count, so the UT actually verifies that the parallel path is taken, rather than only checking the arithmetic result.

---

### DIFF
--- a/third_party/ascend/unittest/pytest_ut/test_parallel.py
+++ b/third_party/ascend/unittest/pytest_ut/test_parallel.py
@@ -19,82 +19,41 @@
 # THE SOFTWARE.
 
 import pytest
+import torch
+import torch_npu
+from torch.testing import assert_close
 
 import triton
 import triton.language as tl
-
-import torch
-import torch_npu
-
-import test_common
 import triton.language.extra.cann.extension as extension
 
 
 @triton.jit
-def triton_add(in_ptr0, in_ptr1, out_ptr0, L: tl.constexpr, M: tl.constexpr, N: tl.constexpr):
-    lblk_idx = tl.arange(0, L)
-    mblk_idx = tl.arange(0, M)
-    nblk_idx = tl.arange(0, N)
-    idx = lblk_idx[:, None, None] * N * M + mblk_idx[None, :, None] * N + nblk_idx[None, None, :]
-    x0 = tl.load(in_ptr0 + idx)
-    x1 = tl.load(in_ptr1 + idx)
-    ret = x0 + x1
+def parallel_kernel(x_ptr, out_ptr, M: tl.constexpr, N: tl.constexpr):
+    # Load the full [M, N] block
+    offs_m = tl.arange(0, M)
+    offs_n = tl.arange(0, N)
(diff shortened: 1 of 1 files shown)


hongziqi (dismissed):

WuTYSFG (approved):

third_party/ascend/unittest/pytest_ut/test_parallel.py
```diff
@@ -19,82 +19,41 @@
...
-    shape = (L, M, N)
-    x0 = test_common.generate_tensor(shape=(L, M, N), dtype=sigtype).npu()
-    x1 = test_common.generate_tensor(shape=(L, M, N), dtype=sigtype).npu()
-    y_ref = x0 + x1 + x1 + x1 + x0 + x0 + x0
-
-    output = torch.zeros(shape, dtype=dtype).npu()
-    h = triton_add[1, 1, 1](x0, x1, output, L, M, N)
-
-    test_common.validate_cmp(sigtype, output, y_ref)
+def parallel_kernel(x_ptr, out_ptr, M: tl.constexpr, N: tl.constexpr):
```
hongziqi, line 32: It is suggested to supplement the PR description with the reason for modifying the test cases.

hongziqi (approved):

PR #681, opened 2026-06-18
[CI](feat) Add docs preview status workflow

xuedinge233:
### Summary

Mirrors the Read the Docs PR build state onto a GitHub commit status so contributors see the docs preview link in the PR checks list without the RTD GitHub App installed.

- Reads build state from the documented project builds list filtered by the PR head commit (GET /builds/?commit=<sha>), which pins the status to the exact commit under test. 
- Treats both "finished" (success flag decides) and "cancelled" as terminal states so a cancelled build is reported instead of polling to the timeout.
- RTD_TOKEN is optional: a public project's builds API is read anonymously; the token is only sent when configured (private project or higher rate limit).
- Polls every 30s for up to 20 min (~40 requests/run), well under anonymous rate limits.

Runs via pull_request_target so the built-in GITHUB_TOKEN can write commit statuses; no PR code is checked out.

### DIFF
--- a/.github/workflows/docs-preview-status.yml
+++ b/.github/workflows/docs-preview-status.yml
@@ -1,0 +1,136 @@
+name: Docs preview status
+
+# Mirrors the RTD PR build state onto a GitHub commit status, so contributors
+# see the docs preview link in the PR checks list without needing the RTD
+# GitHub App installed.
+#
+# Optional repo secret: RTD_TOKEN (RTD account -> API tokens). Only needed for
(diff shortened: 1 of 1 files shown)


xuedinge233:
The current task has not been merged into the main trunk, so the task flow cannot be seen in the checklist yet. We can see the [docs on the rtd here](https://triton-ascend--681.org.readthedocs.build/)

xuedinge233:
The following is the test record of the fork warehouse
https://github.com/xuedinge233/triton-ascend/pull/39

PR #732, opened 2026-06-23
[launcher](fix) fix kernel prof name

zhanwei33:
# New contributor declaration

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests

# Fix stale Ascend profiler kernel names

## Summary

This change fixes incorrect kernel names in msprof output for Ascend Triton launches.

The generated NPU launcher previously stored `kernelName` and `tensorKinds` as static C++ state. When multiple Triton kernels shared the same generated launcher module, the first launch initialized those static values and later launches reused them. As a result, profiler records could report the first kernel name instead of the actual kernel being launched.

The launcher now reads `packedMetadata["kernel_name"]` and `packedMetadata["tensor_kinds"]` on every Python launcher call and passes those per-launch values into `_launch()`.

## Changes

- Remove static `kernelName` and `tensorKinds` from the generated Ascend launcher source.
- Rebuild `kernelName` and `tensorKinds` from `packedMetadata` for each launch.
- Add a regression test that verifies generated launcher source does not cache the kernel name and reads it from `packedMetadata` per launch.
- Override the NPU assignment fixture in this pure source-generation test file so the test can run without requiring an actual NPU device.

## Impact

This change affects profiler metadata reporting only. It does not change kernel binary loading, argument packing, grid configuration, workspace allocation, taskqueue behavior, or runtime function handles.

Profiler switch state (`__MsprofFlagL0` / `__MsprofFlagL1`) remains static as before. Only per-kernel metadata that can vary between launches is moved back to local per-launch state.

### DIFF
--- a/third_party/ascend/backend/driver.py
+++ b/third_party/ascend/backend/driver.py
@@ -611,8 +611,6 @@ extern "C" {
   extern int MsprofRegisterCallback(unsigned int moduleId, callback handle);
   static unsigned int __MsprofFlagL0  = 0;
   static unsigned int __MsprofFlagL1  = 0;
-  static const char* kernelName = nullptr ;
-  static std::vector<int> tensorKinds;
 
   int ProfCtrlHandle(unsigned int CtrlType, void* CtrlData, unsigned int DataLen) {
     if ((CtrlData == nullptr) || (DataLen == 0U)) {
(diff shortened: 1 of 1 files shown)


hongziqi (commented):
It is recommended to supplement the PR description.

third_party/ascend/backend/driver.py
```diff
@@ -1099,20 +1097,17 @@ def _format_of_msprof_task_type_ratio(bs_task_type, mix_mode):
 
 
   // get kernel_name
-  if (!kernelName) {{
-      PyObject *kernelNameObj = PyDict_GetItemString(packedMetadata, "kernel_name");
-      kernelName = PyUnicode_AsUTF8(kernelNameObj);
-  }}
+  PyObject *kernelNameObj = PyDict_GetItemString(packedMetadata, "kernel_name");
```
hongziqi, line 1100: It is recommended to supplement the PR description.

hongziqi (approved):

WuTYSFG (approved):

PR #740, opened 2026-06-23
[release/3.2.2](fix) Fix boundary size calculation for make_block_ptr

hongziqi:
## Summary
This PR fixes **incorrect boundary_check lowering** for block pointer paths by reworking the core computation strategy.
The key change: **prefer direct logical-axis boundary size calculation using traced tensor pointer metadata** as the main path, while keeping the linear offset-based method as a safe fallback. The fallback path’s partial boundary_check bug is also fixed.
## Problem
`boundary_check` is a **block-pointer-native semantic** that requires per-axis valid element counts (`boundarySizes`):

$$
boundarySizes = [size_0, size_1, \dots, size_{n-1}]
$$

where each entry says how many elements are still valid on that logical axis.

The previous implementation suffered from **two critical flaws**:
### 1. Wrong Primary Abstraction
Block pointers carry full logical-axis metadata (shape, offsets, block shape) by design.
However, the old code **overused linear offset reconstruction** (via div/rem + memref strides) as the primary way to recover per-axis offsets.
This method is **not generally correct** for block pointers, especially with implicit permute or non-standard layouts.
### 2. Partial Boundary Check Bug in getBoundarySizes

`getBoundarySizes` incorrectly stripped `offsetShift` **only on boundary-checked axes**, allowing unchecked axes to pollute checked axes.

This caused wrong out-of-bounds detection and excessive zero-padding.
### Example 1: Partial boundary_check
Consider a 2D row-major tensor:
- `shape = (32, 50)`
- `strides = (50, 1)`
- `block_shape = (8, 8)`
- `offsets = (10, 40)`
- `boundary_check = (1,)`

The flattened linear offset is:

$$
linear\_offset = offset_0 \times stride_0 + offset_1 \times stride_1 = 10 \times 50 + 40 \times 1 = 540
$$

Old fallback incorrectly computes axis 1 offset as:

$$
curOffset_1 = \frac{linear\_offset}{stride_1} = \frac{540}{1} = 540
$$

What it should do is peel off the higher-axis contribution before recovering axis 1:

$$
curOffset_1 = \frac{linear\_offset \bmod stride_0}{stride_1} = \frac{540 \bmod 50}{1} = 40
$$
### Example 2: Full boundary_check on a real block pointer

Consider a 3D block pointer with logical metadata:
- `shape = (17, 13, 11)`
- `offsets = (12, 10, 9)`
- `block_shape = (8, 4, 4)`
- `boundary_check = (0, 1, 2)`

>The old design still tended to route this through linear offset reconstruction from layout information. That is fragile because full-axis validity is fundamentally a logical-axis question, not a memref-layout question.

The new primary path computes each checked axis directly:

$$
leftSize_i = \max(shape_i - offset_i, 0)  \qquad
boundarySize_i = \min(blockShape_i, leftSize_i)
$$

Substituting values:
- axis 0: `leftSize_0 = max(17 - 12, 0) = 5`, so `boundarySize_0 = min(8, 5) = 5`
- axis 1: `leftSize_1 = max(13 - 10, 0) = 3`, so `boundarySize_1 = min(4, 3) = 3`
- axis 2: `leftSize_2 = max(11 - 9, 0) = 2`, so `boundarySize_2 = min(4, 2) = 2`

Therefore:

$$
boundarySizes = [5, 3, 2]
$$
## Solution

When logical-axis metadata can be traced:
1. Use `traceTensorPtrAxisInfo` to recover `shape` and `offsets`
2. Use `getBoundarySizesForTensorPtrInfo` to compute valid sizes directly
   Formula: $$boundarySize_i = min(blockShape_i, max(shape_i - offset_i, 0))$$
If metadata is lost (e.g., lowered memref view):
- Fall back to `getBoundarySizes`
- Fixed to strip offsetShift across all axes, not just checked ones
### Lowering Workflow

```mermaid
flowchart TD
    A[Load/Store with boundary_check] --> B[Trace tensor-ptr axis info]
    B -->|success| C[Compute boundarySizes from logical shape/offsets]
    B -->|fail| D[Fallback to memref-based getBoundarySizes]
    C --> E[Create subview or slice with boundarySizes]
    D --> E
    E --> F[Lower copy or materialize_in_destination]
```
## CheckList

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests
  - [x] This PR does not need a test because `FILL THIS IN`.

### DIFF
--- a/third_party/ascend/lib/TritonToLinalg/LoadStoreConverter.cpp
+++ b/third_party/ascend/lib/TritonToLinalg/LoadStoreConverter.cpp
@@ -344,16 +344,16 @@ LoadConverter::matchAndRewrite(triton::LoadOp op, OpAdaptor adaptor,
   // boundary check
   auto boundaryCheck = op.getBoundaryCheck();
   if (!boundaryCheck.empty()) {
-    auto makeTensorPtrOp = op.getPtr().getDefiningOp<triton::MakeTensorPtrOp>();
-    auto boundarySizes = mlir::ConverterUtils::getBoundarySizes(
-        boundaryCheck, /*remapped*/ ptr, loc, rewriter);
+    std::optional<mlir::ConverterUtils::TensorPtrAxisInfo> tensorPtrInfo;
+    auto boundarySizes = mlir::ConverterUtils::getBoundarySizesFromTensorPtrInfoOrFallback(
+        op.getPtr(), ptr, boundaryCheck, loc, rewriter, tensorPtrInfo);
     // handle the padding
     auto padding = op.getPadding();
     SmallVector<OpFoldResult> srcOffsets(boundarySizes.size(), rewriter.getIndexAttr(0));
     SmallVector<OpFoldResult> dstOffsets;
-    if (makeTensorPtrOp) {
+    if (tensorPtrInfo) {
       auto zeroVal = rewriter.createOrFold<arith::ConstantOp>(loc, rewriter.getI32IntegerAttr(0));
-      for (auto [idx, offVal] : llvm::enumerate(makeTensorPtrOp.getOffsets())) {
+      for (auto [idx, offVal] : llvm::enumerate(tensorPtrInfo->offsets)) {
         if (llvm::find(boundaryCheck, idx) == boundaryCheck.end()) {
           dstOffsets.push_back(srcOffsets[idx]);
           continue;
@@ -1043,14 +1043,14 @@ StoreConverter::matchAndRewrite(triton::StoreOp op, OpAdaptor adaptor,
   // 1. boundary size check
   auto boundaryCheck = op.getBoundaryCheck();
   if (!boundaryCheck.empty()) {
-    auto makeTensorPtrOp = op.getPtr().getDefiningOp<triton::MakeTensorPtrOp>();
-    auto boundarySizes = mlir::ConverterUtils::getBoundarySizes(
-        boundaryCheck, /*remapped*/ ptr, loc, rewriter);
+    std::optional<mlir::ConverterUtils::TensorPtrAxisInfo> tensorPtrInfo;
+    auto boundarySizes = mlir::ConverterUtils::getBoundarySizesFromTensorPtrInfoOrFallback(
+        op.getPtr(), ptr, boundaryCheck, loc, rewriter, tensorPtrInfo);
     SmallVector<OpFoldResult> srcOffsets;
     SmallVector<OpFoldResult> dstOffsets(boundarySizes.size(), rewriter.getIndexAttr(0));
-    if (makeTensorPtrOp) {
+    if (tensorPtrInfo) {
       auto zeroVal = rewriter.createOrFold<arith::ConstantOp>(loc, rewriter.getI32IntegerAttr(0));
-      for (auto [idx, offVal] : llvm::enumerate(makeTensorPtrOp.getOffsets())) {
+      for (auto [idx, offVal] : llvm::enumerate(tensorPtrInfo->offsets)) {
         if (llvm::find(boundaryCheck, idx) == boundaryCheck.end()) {
           srcOffsets.push_back(dstOffsets[idx]);
           continue;
--- a/third_party/ascend/lib/TritonToLinalg/LoadStoreConverter.cpp
+++ b/third_party/ascend/lib/TritonToLinalg/LoadStoreConverter.cpp
@@ -1072,13 +1072,35 @@ StoreConverter::matchAndRewrite(triton::StoreOp op, OpAdaptor adaptor,
     } else {
       srcOffsets = dstOffsets;
     }
-    auto srcSlice = mlir::ConverterUtils::makeExtractSliceOp(
-        val, srcOffsets, boundarySizes, loc, rewriter);
     auto dstSubview = mlir::ConverterUtils::makeSubViewOp(
         ptr, dstOffsets, boundarySizes, loc, rewriter);
-    auto storeOp = rewriter.create<bufferization::MaterializeInDestinationOp>(
-        loc, srcSlice, dstSubview);
-    storeOp.setWritable(true);
+
+    bool isCheckedAxisPrefix = !boundaryCheck.empty();
+    for (auto [expectedAxis, checkedAxis] : llvm::enumerate(boundaryCheck)) {
+      if (checkedAxis != static_cast<int32_t>(expectedAxis)) {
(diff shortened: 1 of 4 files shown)


third_party/ascend/lib/Utils/Utils.cpp
```diff
@@ -395,22 +531,29 @@ getBoundarySizes(llvm::ArrayRef<int32_t> boundaryCheck, Value ptr,
...
+    OpFoldResult curStride = fullShapeReCast.getConstifiedMixedStrides()[i];
     if (llvm::find(boundaryCheck, i) != boundaryCheck.end()) {
-      auto fullShape = fullShapeReCast.getConstifiedMixedSizes()[i];
+      if (isZero(curStride)) {
+        emitWarning(loc)
+            << "getBoundarySizes() cannot reconstruct boundary on checked "
+               "zero-stride axis "
+            << i << "; keep current block size for this axis";
+        continue;
+      }
```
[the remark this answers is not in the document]
hongziqi: Already update

third_party/ascend/lib/Utils/Utils.cpp
```diff
@@ -320,6 +320,142 @@ std::optional<Operation *> getFullShapeOp(Value val,
...
+
+  if (auto blockArg = dyn_cast<BlockArgument>(ptr)) {
+    if (auto forOp = dyn_cast_or_null<scf::ForOp>(
+            blockArg.getOwner()->getParentOp())) {
+      auto init = forOp.getTiedLoopInit(blockArg);
+      if (!init)
+        return std::nullopt;
+      return traceTensorPtrAxisInfo(init->get(), rewriter, loc);
+    }
+  }
```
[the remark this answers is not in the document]
hongziqi: already update for this

WuTYSFG (dismissed):

third_party/ascend/lib/Utils/Utils.cpp
```diff
@@ -325,6 +320,184 @@ std::optional<Operation *> getFullShapeOp(Value val,
...
+      info.shape = baseInfo->shape;
+      info.offsets.reserve(baseInfo->offsets.size());
+      for (auto [baseOffset, advanceOffset] :
+           llvm::zip(baseInfo->offsets, advanceOp.getOffsets())) {
+        Value iterCountTyped = iterCount;
+        if (iterCountTyped.getType() != advanceOffset.getType()) {
+          iterCountTyped = rewriter.createOrFold<arith::IndexCastOp>(
+              loc, advanceOffset.getType(), iterCountTyped);
+        }
+        Value totalAdvance = rewriter.createOrFold<arith::MulIOp>(
```
KanuaK, line 444: Can this handle the case where `advanceOffset` is different for each loop iteration?

third_party/ascend/lib/TritonToLinalg/LoadStoreConverter.cpp
```diff
@@ -1072,13 +1072,57 @@ StoreConverter::matchAndRewrite(triton::StoreOp op, OpAdaptor adaptor,
...
+      auto srcSubview = mlir::ConverterUtils::makeSubViewOp(
+          valueMemRef, srcOffsets, boundarySizes, loc, rewriter);
+      rewriter.create<memref::CopyOp>(loc, srcSubview, dstSubview);
+    } else {
+      auto srcSlice = mlir::ConverterUtils::makeExtractSliceOp(
+          val, srcOffsets, boundarySizes, loc, rewriter);
+      auto storeOp = rewriter.create<bufferization::MaterializeInDestinationOp>(
+          loc, srcSlice, dstSubview);
+      storeOp.setWritable(true);
+    }
```
KanuaK, line 1125: Better to add an MLIR test to cover this case.

third_party/ascend/lib/Utils/Utils.cpp
hongziqi: Resolved. When advance uses loop variables, the fallback path will be taken.

third_party/ascend/lib/TritonToLinalg/LoadStoreConverter.cpp
hongziqi: The related MLIR test cases will be submitted later. Currently, there are related end-to-end test cases in the library.

third_party/ascend/lib/TritonToLinalg/LoadStoreConverter.cpp
```diff
@@ -1043,14 +1043,14 @@ StoreConverter::matchAndRewrite(triton::StoreOp op, OpAdaptor adaptor,
   // 1. boundary size check
   auto boundaryCheck = op.getBoundaryCheck();
   if (!boundaryCheck.empty()) {
-    auto makeTensorPtrOp = op.getPtr().getDefiningOp<triton::MakeTensorPtrOp>();
-    auto boundarySizes = mlir::ConverterUtils::getBoundarySizes(
-        boundaryCheck, /*remapped*/ ptr, loc, rewriter);
+    std::optional<mlir::ConverterUtils::TensorPtrAxisInfo> tensorPtrInfo;
+    auto boundarySizes = mlir::ConverterUtils::getBoundarySizesFromTensorPtrInfoOrFallback(
```
hongziqi, line 1047: rename boundarySizes to realOffsets

third_party/ascend/lib/Utils/Utils.cpp
```diff
@@ -325,6 +320,210 @@ std::optional<Operation *> getFullShapeOp(Value val,
...
+// Control flow:
+//
+//   ptr
+//    |
+//    +-- MakeTensorPtrOp --------> return {shape, offsets}
+//    |
+//    +-- AdvanceOp --------------> trace(base)
+//    |                               then offsets := baseOffsets + advance
+//    |
+//    +-- scf.for BlockArgument --> trace(tied loop init), or if yield is
```
hongziqi, line 351: update offset in iter val

KanuaK (approved):

third_party/ascend/lib/TritonToLinalg/LoadStoreConverter.cpp
hongziqi: Track in issue 804.

third_party/ascend/lib/Utils/Utils.cpp
hongziqi: Track in issue 804.

kaixin1976 (approved):

PR #741, opened 2026-06-23
[CI](fix) Modify the A5 pipeline starter

xuedinge233:
Due to compliance issues, the A5 pipeline will be started here using a public account.


xuedinge233:
The pre-commit check error comes from the previously submitted pr and has nothing to do with this pr.

hongziqi (approved):

HinPeng (approved):

PR #747, opened 2026-06-23
[CompileOps](feat) Revise superblock factor value

darkbuck:
- Now the factor is no longer the shift of the factor value but the real factor value. Change its initial value to 1 and add superblock factor option when it's greater than 1.

# New contributor declaration

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests

### DIFF
--- a/third_party/ascend/backend/compiler.py
+++ b/third_party/ascend/backend/compiler.py
@@ -988,7 +988,7 @@ class NPUOptions:
     disable_fma: bool = False
 
     # superblocking factor
-    superblock_factor: int = 0
+    superblock_factor: int = 1
 
     def __post_init__(self):
         # Parse compile_mode and set related fields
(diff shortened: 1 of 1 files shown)


CalvonBai:
retry

third_party/ascend/backend/compiler.py
```diff
@@ -988,7 +988,7 @@ class NPUOptions:
     disable_fma: bool = False
 
     # superblocking factor
-    superblock_factor: int = 0
+    superblock_factor: int = 1
```
hongziqi, line 991: Can you briefly explain the usage and introduction of this compilation option?

darkbuck: We change that factor from the previous shift of the real value to the real value to keep it consistent with other options and allow a non-power-of-2 factor later. The default value is changed from 0 to 1 accordingly.

darkbuck: Factor 1 means the superblocking is disabled.

KanuaK (approved):

third_party/ascend/backend/compiler.py
```diff
@@ -1063,7 +1063,7 @@ def ttir_to_npubin(mod, metadata, opt):
             # cap keys off the same env switch, so the two stay in sync.
             if _is_auto_map_parallel_blocks_enabled():
                 _compile_option_list += ["--enable-auto-blockify-loop"]
-                if opt.superblock_factor > 0:
+                if opt.superblock_factor > 1:
```
hongziqi, line 1066: The default behavior has not been changed. If changed to >1

hongziqi (approved):

KanuaK:
> /lgtm

@hailijuan You could give review conclusion of `Approve` in `FileChanged` Page

PR #790, opened 2026-06-24
[simt](feat) the parsing priority of simt_stack_limit is adjusted

LH-123L:
simt_stack_limit resolution precedence: 
1.torch_npu's acl_default.json "StackSize":{"simt_stack_size":N} takes precedence and the user-specified value is ignored.
2.if that config key is absent ,fail back to the kernel-time user-specified simt_stack_limit
https://wiki.huawei.com/domains/148244/wiki/296387/WIKI2026052611224469
By default, simt_stack_size is obtained from the PTA. After the configuration in acl_default.json is modified, you do not need to configure simt_stack_limit. The system ensures that the configured value is consistent with the verified data, which is more user-friendly and reduces user operations.

Modify document at  https://github.com/triton-lang/triton-ascend/pull/853

# New contributor declaration

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests

### DIFF
--- a/third_party/ascend/backend/compiler.py
+++ b/third_party/ascend/backend/compiler.py
@@ -22,6 +22,7 @@ import ctypes
 import functools
 import hashlib
 import glob
+import json
 import os
 import re
 import subprocess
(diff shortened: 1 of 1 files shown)


python/setup.py
```diff
@@ -828,7 +828,7 @@ def get_default_version():
     version_file = Path(__file__).parent / "version.txt"
     if version_file.exists():
         return version_file.read_text().strip()
-    return "3.2.2"
+    return "3.2.2-dev"
```
LH-123L: 打包时whl包名与release/3.2.2重复，为做区分修改dev分支版本号

KanuaK: 是否应该把调整版本号与simt stack size的调整分成两个pr？便于dev与main同步时隔离commit

PR #815, opened 2026-06-25
[libdevice](revert) Revert high-performance optimization scheme.

XMU-qcj:
# New contributor declaration

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests

[libdevice](revert) Revert libdevice

These two PRs involve the following:

The default entry now uses the libdevice high‑performance solution, which has caused some test cases for libdevice.tanh to fail.

Optimize the performance of lgamma. Together with the former PR, both are being reverted.


KanuaK (approved):

third_party/ascend/backend/compiler.py
```diff
@@ -50,6 +50,7 @@
     _warn_auto_blockify_disabled,
     downgrade_llir,
     force_disable_ffts,
+    triton_enable_libdevice_simt,
```
hongziqi, line 53: Could you elaborate on the background and reasons for this revert? It is recommended to add these details to the PR description.

XMU-qcj: The modifications introduced by these two PRs to libdevice caused some test cases for libdevice.tanh to fail. We would now like to revert to a fully functional state

hongziqi (approved):

hongziqi:
It is recommended to optimize the PR title, as the current title is somewhat vague.

PR #835, opened 2026-06-27
[MaskAnalysis](fix) extend clampToNonNegativeIndex to dynamic dims fo…

CHNJZ:
## Problem
`MaskState::clampToNonNegativeIndex` only lower-bounded **constant** dims; non-constant
(dynamic) dims were returned unmodified. In `minStates` the intersection of two independent
masks can be disjoint, producing a negative extent that corrupts the load/store region.

## Per-call-site analysis
Every dim has `end >= start`. Whether `newDim` can go negative depends on whether both
operands of the subtraction are clamped into `[start, end]`.

| Site | Context | `newDim` formula | Non-negative by construction | Needs dynamic clamp |
|------|---------|------------------|------------------------------|---------------------|
| minStates | intersection (also via parseAnd) | `min(lhsEnd,rhsEnd) − max(lhsOff,rhsOff)` | No — disjoint ranges | Yes |
| parseCmp slt | `lhs < rhs` | `min(end, max(start,scalar)) − start` | Yes — both terms ≥ start | No (redundant) |
| parseCmp sle | `lhs <= rhs` | `min(end, max(start,rhs+1)) − start` | Yes | No (redundant) |
| parseCmp sge | `lhs >= rhs` | `end − min(end, max(start,scalar))` | Yes — newStart ∈ [start,end] | No (redundant) |

The cmp paths already lower-bound via `max(start,…)`/`min(end,…)`, so only `minStates`
can yield a negative dim.

## Fix
Gate a dynamic `max(value, 0)` (via `maxOpFoldResult`) behind a `clampDynamic` flag,
enabled only at the `minStates` call site. The cmp paths keep prior behavior, avoiding the
atomic max/min UT regression that motivated the original const-only clamp.

## Test
`pytest_ut/test_mask_negative_dim_clamp.py`: three-segment cat + indirect int64 indexing
repro whose compound masks go through `parseAnd → minStates`, asserted against an eager
reference.

### DIFF
--- a/third_party/ascend/unittest/pytest_ut/test_mask_negative_dim_clamp.py
+++ b/third_party/ascend/unittest/pytest_ut/test_mask_negative_dim_clamp.py
@@ -1,0 +1,60 @@
+import torch
+import triton
+import triton.language as tl
+
+
+@triton.jit
+def kernel_minimal(
+    idx_ptr, table_ptr, out_ptr,
+    y0_numel, x1_numel,
+    Y0BLOCK: tl.constexpr, Y0BLOCK_SUB: tl.constexpr, X1BLOCK_SUB: tl.constexpr,
+):
+    y0_offset = tl.program_id(0) * Y0BLOCK
+    base_y0 = tl.arange(0, Y0BLOCK_SUB)
+    loops_y0 = (Y0BLOCK + Y0BLOCK_SUB - 1) // Y0BLOCK_SUB
+    base_x1 = tl.arange(0, X1BLOCK_SUB)
+    loops_x1 = (x1_numel + X1BLOCK_SUB - 1) // X1BLOCK_SUB
+    for loop_y0 in range(loops_y0):
+        y0 = y0_offset + (loop_y0 * Y0BLOCK_SUB) + base_y0[:, None]
+        y0_mask = y0 < min(Y0BLOCK + y0_offset, y0_numel)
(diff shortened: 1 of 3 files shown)


third_party/ascend/unittest/pytest_ut/test_mask_negative_dim_clamp.py
```diff
@@ -0,0 +1,69 @@
...
+        y0_mask = y0 < min(Y0BLOCK + y0_offset, y0_numel)
+        for loop_x1 in range(loops_x1):
+            x1 = (loop_x1 * X1BLOCK_SUB) + base_x1[None, :]
+            x1_mask = x1 < x1_numel
+
+            tmp27 = (y0 >= 64) & (y0 < 128) & y0_mask
+            tmp31 = tl.load(in_ptr2 + ((-64) + y0), tmp27, other=0)
+
+            tmp48 = (y0 >= 128) & y0_mask
+            tmp52 = tl.load(in_ptr2 + ((-128) + y0), tmp48, other=0)
```
WuTYSFG: Suggest simplifying the test case

CHNJZ: done

KanuaK (approved):

WuTYSFG (approved):

PR #916, opened 2026-07-06
[init](feat) Eliminate 1.15s cold-start overhead of `get_ascend_devices`

kingsirvince:
# PR: Eliminate 1.15s cold-start overhead of `get_ascend_devices`

## Background

During multi-process concurrent cold start of `import torch; import torch_npu`, the `triton.tools.get_ascend_devices` module took **1.157s**, making it the dominant bottleneck in the triton import chain.

Root cause: [get_ascend_devices.py](python/triton/tools/get_ascend_devices.py) eagerly executes `check_npu_smi_device()` at module top level. This function invokes the `npu-smi info` command via `subprocess` (second-level latency). Importing the module triggers it, even when the value is never used.

### 1. Slow operation executed at module top level

```python
# Before: executed on import
ascend_devices = get_ascend_devices()
npu_smi_condition = check_npu_smi_device()   # ← npu-smi info, second-level
is_compile_on_910_95 = pci_condition or npu_smi_condition
```

### 2. Eager evaluations in the cold-start path defeat lazy loading

After lazy-loading only `get_ascend_devices.py`, the 1.15s overhead simply moved from `get_ascend_devices` to `triton.backends` (1.18s → 1.14s, essentially unchanged). Reason:

- `triton/backends/__init__.py`'s `_discover_backends()` dynamically loads `ascend/compiler.py`
- `compiler.py` / `autotuner.py` use `from triton.tools.get_ascend_devices import is_compile_on_910_95` — **`from X import Y` evaluates Y immediately**, triggering the lazy `__getattr__`
- `compiler.py` also uses `is_compile_on_910_95` as a dataclass default — **evaluated at class definition time**

## Optimization Approach

Core idea: **defer the `npu-smi info` execution from import time to first-use time (kernel compile time)**, with no eager evaluation anywhere in the cold-start chain.

### Changes

| File | Change | Rationale |
|---|---|---|
| [get_ascend_devices.py](python/triton/tools/get_ascend_devices.py) | Replace top-level eager computation with PEP 562 `__getattr__` lazy loading + cache | Import no longer triggers npu-smi; first attribute access triggers it and caches the result |
| [compiler.py](third_party/ascend/backend/compiler.py) | `from ... import is_compile_on_910_95` → `from triton.tools import get_ascend_devices`; dataclass defaults changed to `field(default_factory=...)` | Module import doesn't trigger; evaluated only at dataclass instantiation |
| [autotuner.py](third_party/ascend/backend/runtime/autotuner.py) | Same module-import rewrite + class attribute access moved to `__init__` | Deferred to Autotuner instantiation |
| [utils.py](third_party/ascend/backend/utils.py) | Module import + qualified access inside functions | Evaluated only on function call |
| [semantic.py](python/triton/language/semantic.py) | Same as above | Evaluated only on function call |
| [vec_ops.py](third_party/ascend/language/cann/extension/vec_ops.py) | Same as above | Evaluated only on function call |
| [language/\_\_init\_\_.py](python/triton/language/__init__.py), [extension/\_\_init\_\_.py](third_party/ascend/language/cann/extension/__init__.py) | Removed re-export imports | No direct imports in the cold-start path |

### Key Technical Points

1. **PEP 562 `__getattr__`**: Triggered on module attribute lookup failure, enabling lazy loading
2. **`from X import Y` vs `from X import m`**: The former evaluates Y immediately; the latter only fetches a module reference
3. **dataclass `field(default_factory=...)`**: Default value evaluated at instantiation time rather than class definition time

## Results

| Metric | Before | After | Improvement |
|---|---|---|---|
| Total time (`import torch + torch_npu`) | 9.362s | 8.522s | **-8.9%** |
| triton import time | 1.286s | 0.134s | **-89.6%** |
| `get_ascend_devices` module time | 1.157s | 0.222ms | **-99.98%** |
| `triton.backends` module time | 1.176s | 0.017s | **-98.6%** |

## Compatibility

- External API unchanged: `is_compile_on_910_95` remains accessible with the same value as before
- Only behavioral change: `npu-smi info` first executes at first access of `is_compile_on_910_95` (i.e., kernel compile time) instead of at import time
- Test files untouched (their `from ... import is_compile_on_910_95` triggers lazy loading at test runtime, which works correctly)


hongziqi (dismissed):

python/triton/tools/get_ascend_devices.py
```diff
@@ -49,7 +49,30 @@ def check_npu_smi_device():
...
+    _lazy_cache.update({
+        'ascend_devices': ascend_devices,
+        'pci_condition': pci_condition,
+        'npu_smi_condition': npu_smi_condition,
+        'is_compile_on_910_95': is_compile_on_910_95,
+    })
+
+
+def __getattr__(name):
+    if name in ('ascend_devices', 'pci_condition', 'npu_smi_condition', 'is_compile_on_910_95'):
```
WuTYSFG: 这个地方需要支持这么多attr吗

WuTYSFG (dismissed):

python/triton/tools/get_ascend_devices.py
```diff
@@ -49,7 +49,30 @@ def check_npu_smi_device():
         return False
 
 
-ascend_devices = get_ascend_devices()
-pci_condition = any("0xd806" in dev for dev in ascend_devices)
-npu_smi_condition = check_npu_smi_device()
-is_compile_on_910_95 = pci_condition or npu_smi_condition
\ No newline at end of file
+# Lazy module attributes (PEP 562). The values below are computed on first
```
KanuaK: Could we also consider applying this lazy-access pattern to other Python modules that still import `is_compile_on_910_95` directly, such as `third_party/ascend/backend/compiler.py` and `third_party/ascend/backend/runtime/autotuner.py`?

kingsirvince: If other modules take a long time to import, you can also use the lazy_init .  However, currently, only this command is observed to take a long time.

python/triton/tools/get_ascend_devices.py
kingsirvince: 这里将会改进，只暴露is_compile_on_910_95，删除其他中间变量暴露。

PR #917, opened 2026-07-07
[build](fix) support Ascend-only runtime builds

ShuhaoZhangTony:
### Summary

Make Ascend-only Triton runtime builds tolerate optional AMD/NVIDIA/Proton/test components that may be unavailable or intentionally disabled in an Ascend-focused source checkout.

This PR is extracted from an Ascend NPU runtime bring-up where importing/building the Ascend path could still discover or link backend-specific components outside the intended Ascend runtime surface.

### Changes

- Avoid registering unavailable AMD/NVIDIA/NVWS/test passes and dialects in Ascend-only tool registration paths.
- Keep proton plugin symbols out of the Python extension when proton itself is disabled.
- Add guards around optional backend-specific CMake targets.
- Skip missing Python backend entry points during backend discovery while still re-raising unrelated import errors.
- Fix dependent-template compilation issues in Ascend `BlockPtrAnalysis`.
- Allow the intended `createMemAccOp` specialization pattern in the unstructure conversion pass.

### Validation

- `PYTHONDONTWRITEBYTECODE=1 python3 -B` AST parse of `python/triton/backends/__init__.py`
- `git diff --check`
- The same patch was validated in the vLLM-HUST Ascend-only runtime source checkout before being proposed upstream.

### Notes for reviewers

This touches build registration and optional backend linkage, so CI coverage across non-Ascend paths is important. The goal is not to remove multi-backend support, but to avoid requiring disabled/unavailable backend components for an Ascend-only runtime build/import path.


ShuhaoZhangTony:
This broad PR has been superseded by the focused upstream PRs #918, #919, #920, #922, and #923. Those branches are independently reviewable; #922 and #923 have also been rebased onto current main. Closing this duplicate avoids maintaining the same fixes in two conflicting branches.

PR #918, opened 2026-07-07
[runtime](fix) skip missing backend entry points

ShuhaoZhangTony:
## Summary
- skip backend entry points whose backend package is not importable
- keep raising `ModuleNotFoundError` for missing nested dependencies inside an importable backend
- add a regression test for a stale/missing backend entry point

## Why
A source checkout or environment can contain entry point metadata for a backend package that is not present on `PYTHONPATH`. In that case backend discovery currently fails the whole `triton.backends` import. Skipping only the entry point whose package is absent lets the remaining installed backends continue to load while preserving real dependency failures inside installed backends.

## Validation
- `PYTHONDONTWRITEBYTECODE=1 python3 - <<PY ... ast.parse(...) ... PY` on the changed files
- `git diff --check`

`PYTHONPATH=python python3 -m pytest python/test/unit/runtime/test_backend_discovery.py -q` could not run in my local checkout because the existing local `python/triton/_C/libtriton.so` requires `GLIBC_2.38`, which is newer than this host glibc.

### DIFF
--- a/python/test/unit/runtime/test_backend_discovery.py
+++ b/python/test/unit/runtime/test_backend_discovery.py
@@ -1,0 +1,25 @@
+from types import SimpleNamespace
+
+import triton.backends as triton_backends
+
+
+class _EntryPoints:
+
+    def __init__(self, entries):
+        self.entries = entries
+
+    def select(self, group):
+        assert group == "triton.backends"
+        return self.entries
+
+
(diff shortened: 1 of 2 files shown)


ShuhaoZhangTony:
Gentle ping for code-owner review. The wheel build and Ascend integration tests are green; this PR is ready for maintainer feedback.

PR #921, opened 2026-07-07
[build](fix) drop test libs from production tools

ShuhaoZhangTony:
## Summary
- remove test-only Triton libraries from production tool link lines
- keep `triton-opt`, `triton-reduce`, `triton-lsp`, and `triton-tensor-layout` independent of test targets

## Why
Production tools should not require test dialect/analysis targets to be built. This helps minimal or Ascend-focused builds that do not materialize test libraries.

## Validation
- `git diff --check`
- verified `bin/CMakeLists.txt` no longer references `TritonTestAnalysis`, `TritonTestDialect`, `TritonAMDGPUTestAnalysis`, or `TritonTestProton`

### DIFF
--- a/bin/CMakeLists.txt
+++ b/bin/CMakeLists.txt
@@ -30,11 +30,6 @@ target_link_libraries(triton-opt PRIVATE
   BiShengIRScopeDialect
   BiShengIRSymbolDialect
   BiShengIRTensorDialect
-  # tests
-  TritonTestAnalysis
-  TritonTestDialect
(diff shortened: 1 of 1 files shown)


ShuhaoZhangTony:
Closing this split because it is not independently buildable: dropping the test libraries from the production tools leaves test-pass registrations in RegisterTritonDialects.h, which produces unresolved mlir::test::* symbols. The independently buildable cleanup is tracked in #923, which trims the optional test dialect/pass registrations directly.

PR #922, opened 2026-07-07
[build](fix) fix Python module build dependencies

ShuhaoZhangTony:
## Summary
- only append the proton plugin to the Python extension when `TRITON_BUILD_PROTON` is enabled
- avoid linking selected NVIDIA/AMD lowering libraries into the Python extension library set
- materialize the AMDGPU IR/tablegen targets when the Python module build needs them

## Why
Minimal Python-module builds can fail when optional profiler/plugin components or AMD tablegen targets are not materialized in the expected order. This keeps optional plugin symbols gated while adding the explicit AMD target dependencies needed by the extension build.

## Validation
- `git diff --check`
- local textual check for the new CMake target references

The authoritative validation is the repository wheel/build CI because this patch changes CMake target wiring.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -246,6 +246,10 @@ if(TRITON_BUILD_PYTHON_MODULE)
 
   get_property(triton_libs GLOBAL PROPERTY TRITON_LIBS)
   get_property(triton_plugins GLOBAL PROPERTY TRITON_PLUGINS)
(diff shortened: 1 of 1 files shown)


ShuhaoZhangTony:
Gentle ping for code-owner review. The wheel build, code-review job, and Ascend integration tests are green; this PR is ready for maintainer feedback.

PR #924, opened 2026-07-07
[build](fix) materialize NVWS transform targets

ShuhaoZhangTony:
## Summary
- explicitly add NVWS IR and transform subdirectories when `TritonGPUTransforms` needs those targets
- guard the additions with `NOT TARGET` checks so existing full builds keep their target ownership

## Why
Some build configurations reach `TritonGPUTransforms` without materializing the NVWS targets first. This makes the target dependency ordering explicit while preserving existing full-build behavior.

## Validation
- `git diff --check`

The authoritative validation is the repository CMake/wheel CI.


ShuhaoZhangTony:
Closing this split because it is not independently buildable: materializing the NVWS transform targets this way causes duplicate add_subdirectory use of the NVWS IR/Transforms source directories during wheel configuration. I will revisit this as a smaller CMake fix if needed, but this PR should not be reviewed as a standalone upstream patch.

PR #925, opened 2026-07-07
[build](fix) gate Proton backend dependencies

ShuhaoZhangTony:
## Summary
- build the TritonProton Python plugin only when both the Python module and Proton profiler are enabled
- link Proton NVIDIA/AMD lowering libraries only when the corresponding codegen backend is enabled
- add Proton GPU-to-LLVM backend subdirectories only for enabled backends

## Why
Proton is optional, and its backend-specific lowering libraries should not be required in builds that disable Proton or exclude a backend. This keeps minimal/backend-focused builds from depending on unavailable optional Proton targets.

## Validation
- `git diff --check`

The authoritative validation is the repository CMake/wheel CI.


ShuhaoZhangTony:
Closing this split because the wheel builds but integration imports fail with an unresolved init_triton_proton symbol from libtriton.so. The current dependency gate changes the Proton linkage boundary without updating the Python extension initialization contract, so it should not be reviewed as a standalone upstream patch. I will revisit this as a smaller fix around the module init/linkage boundary if needed.

PR #941, opened 2026-07-07
Gather optimization

dmadej-hw:
# Purpose
The purpose of this optimization is to detect gather-like memory access patters that would otherwise be lowered to  inefficient scalar load loops and replace them with `tl.gather` operations.

Currently, only workloads equivalent to a gather along the last dimension are supported.

The most common pattern targeted by this optimization is the load-load-store sequence.

A simplified example of the transformation is shown below:

Before:
```python
indices = load(idx_ptr + idx_grid)
out = load(src_ptr + indices)
```

After:
```python
indices = load(idx_ptr + idx_grid)
if -len(src_gather_axis) <= min(indices) and max(indices) < len(src_gather_axis):
    src = load(src_ptr + src_grid)
    # normalize indices to positive numbers
    out = gather(src, normalize(indices), axis)
else:
    # fallback
    out = load(src_ptr + indices)
```

# Solution
The pass first determines whether a load operation qualifies for optimization and collects all information required to perform the transformation. If any of the validation step fails, the optimization is skipped.

The pass inserts an additional bounds check for the indices passed to `tl.gather`. During this step, valid negative indices are normalized to their corresponding positive values, ensuring that `tl.gather` always receives indices within the valid non-negative range while preserving the original indexing semantics.

For some workloads (for example, "index select"-like patterns), this optimization provides little or no benefit and should therefore be disabled.
 
## Analysis
Qualify the load operation for optimization and collect necessary data. If any of the steps below fails skip optimization.
1. Identify load operation `load(src_ptr + indices)`.
2. Parse operands using OffsetAnalysis (from TritonToUnstructured pass). This results in creating a map that contains PtrOffsetInfo structs for all operations preceding the load operation. PtrOffsetInfo contains information about rank and structure of operation results.
3. Verify if `(src_ptr + indices)` is structured (it doesn't need optimization).
4. Verify if our case DOES NOT match index select pattern (that includes 1D gather as it is the same as 1D index select).
5. Identify the operation that creates indices.
6. Identify the gather axis. For now only last-axis gather pattern if handled. Other axes will be added in separate PR.
7. Identify the operation that is the base for source shape analysis and determine source shape. The shape is not explicitly passed in the analyzed load operation, so it has to be analyzed from strides (multiplication ops) used for indices grid.
8. Check correctness of the source shape.
9. Identify operations defining row offsets and source pointer splat.

## Rewriting
1. Create if..then..else block, where:
a. Condition: are min and max indices in range `[-len(src_shape[gather_axis]; len(src_shape[gather_axis]))`, 
b. `then` branch: gather logic implementation,
c. `else` branch: fallback to original load operation.
2. Create `then` block:
a. Normalize indices (to positive numbers).
b. Create grid for loading source according to determined source shape.
c. Load source.
d. Perform gather on the last axis.
e. Yield result.
3. Create fallback `else` block:
a. Clone original load operation.
b. Yield result.

# Limitations
Increased UB usage due to loading a whole row of the data tensor.
Currently, only workloads equivalent to a gather along the last dimension are supported.

# Performance results
All benchmarks and performance measurements presented were performed on **Ascend 910B3**.

Across the evaluated benchmarks, the optimization achieves an average **x21.2 speedup** over the unoptimized scalar-load implementation and on average reaches **76% of the performance of `tl.gather`** and **101% of the performance pf `torch.gather`.

Results:
image

A minimal benchmark used to validate both correctness and performance is added below.
[test_gather_optimization.py](https://github.com/user-attachments/files/30009694/test_gather_optimization.py)

# New contributor declaration

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests


dmadej-hw:
Moving to main-dev branch

PR #951, opened 2026-07-09
[TritonToLinalg](fix) Enhance boundary check in tensor descriptor

TecJesh:
### Summary
`DescriptorLoad` / `DescriptorStore` Converter unconditionally called
`getFullBoundaryCheckAttr` to emit a full `boundaryCheck` attribute during
lowering. This triggered the boundary-check path in pointer analysis and
generated redundant `cast` conversion ops in the IR, causing the affected test
cases to fail.
When the block shape fully covers the tensor size, a boundary check is not
needed in the first place.
### Changes
- Add helper `isFullCoverage(blockShape, tensorShape)`:  check if the
  ranks match and every tensor dimension is a compile-time constant equal to the
  corresponding block dimension.
- In `DescriptorLoadConverter` and `DescriptorStoreConverter`, emit an empty
  `boundaryCheck` (`{}`) when the full-coverage case is hit; otherwise keep the
  existing `getFullBoundaryCheckAttr` behavior.
### Effect
- The full-coverage case no longer produces a redundant boundary check and the
  associated `cast` ops, fixing the failing test cases.
- The non-full-coverage case (block cannot cover the tensor, or the size is
  dynamic) is unchanged and still performs a full boundary check.

### Impact
Only affects the boundaryCheck generation logic for descriptor load/store.

# New contributor declaration

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests
  - [x] This PR does not need a test because `test case already exists`.

### DIFF
--- a/third_party/ascend/lib/TritonToLinalg/DescriptorConverter.cpp
+++ b/third_party/ascend/lib/TritonToLinalg/DescriptorConverter.cpp
@@ -42,6 +42,7 @@
 #include "mlir/Dialect/Linalg/IR/Linalg.h"
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
 #include "mlir/Dialect/Utils/ReshapeOpsUtils.h"
+#include "mlir/IR/Matchers.h"
 #include "mlir/IR/OpDefinition.h"
 #include "mlir/IR/ValueRange.h"
 #include "mlir/Transforms/DialectConversion.h"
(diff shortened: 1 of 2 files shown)


third_party/ascend/lib/TritonToLinalg/DescriptorConverter.cpp
```diff
@@ -98,6 +98,22 @@ DenseI32ArrayAttr getFullBoundaryCheckAttr(ConversionPatternRewriter &rewriter,
   return rewriter.getDenseI32ArrayAttr(boundaryCheck);
 }
 
+/// Check if block fully covers the tensor.
+static bool isFullCoverage(ArrayRef<int64_t> blockShape,
```
hongziqi, line 103: It is recommended to optimize the title. The modifications are related to tensor_descriptor, so it's better to mention tensor_descriptor.

TecJesh: done

hongziqi (dismissed):

third_party/ascend/lib/TritonToLinalg/DescriptorConverter.cpp
```diff
@@ -268,7 +284,9 @@ LogicalResult DescriptorLoadConverter::matchAndRewrite(
                                                computeOrder(blockShape) // order
       );
   // 3. replace tt.load
-  auto boundaryCheck = getFullBoundaryCheckAttr(rewriter, blockShape);
+  auto boundaryCheck = isFullCoverage(blockShape, desc.shape)
```
KanuaK: Should this also check the descriptor offsets/indices before dropping `boundaryCheck`? If `blockShape == tensorShape` but the load/store starts from a non-zero offset, the block does not fully cover the tensor from the origin; it may access past the end.

TecJesh: Done

KanuaK (approved):

KanuaK:
lgtm

hongziqi (approved):

PR #953, opened 2026-07-09
[TritonToLinalg](feat) default masked load to zero when `other`/`padd…

WuTYSFG:
…ing` is unset on Ascend

## Summary
The Ascend backend's `TritonToLinalg` pass used to leave the destination buffer uninitialized for masked `triton::LoadOp` when no `other` argument was provided, and for block-pointer loads with `boundary_check` when no `padding_option` was provided. The mask-out / out-of-bound lanes then contained whatever values were in the freshly-allocated memref, which is undefined behavior.
The NVIDIA backend already handles this case by emitting `mov.u{N} %r, 0` as the pre-init of the predicate-load destination (`LoadStoreOpToLLVM.cpp`, `initOperand`). This PR aligns the Ascend backend with that contract.
## Changes
**File:** `third_party/ascend/lib/TritonToLinalg/LoadStoreConverter.cpp`
### 1. Masked load without `other` (around line 480)
Mask path (LoadStoreConverter.cpp): when `mask` is set but `other` is null, build a zero constant of the element type and pass it to `fillTensorWithOtherForMaskScenario` as the fallback value. The subsequent subview + `memref.copy` only overwrites in-bounds lanes,  so the remaining lanes keep the zero fill.
**Before:**
```cpp
if (other) {
  auto scalarOther = mlir::ConverterUtils::getScalarValue(other, loc, rewriter);
  assert(scalarOther && "other value used in masked load ...");
  fillTensorWithOtherForMaskScenario(scalarOther, allocOp, mstate.dims, rewriter);
}
```
**After:**
```
Value effectiveOther;
if (other) {
  effectiveOther = mlir::ConverterUtils::getScalarValue(other, loc, rewriter);
  assert(effectiveOther && "other value used in masked load ...");
} else {
  effectiveOther = rewriter.create<arith::ConstantOp>(loc, rewriter.getZeroAttr(memRefElementType));
}
fillTensorWithOtherForMaskScenario(effectiveOther, allocOp, mstate.dims, rewriter);
```
### 2. Block-pointer load with boundary_check but no padding (around line 404)
Boundary-check path (LoadStoreConverter.cpp): drop the `if (padding.has_value())` guard and use `padding.value_or(PAD_ZERO) == PAD_NAN` to choose the fill value. Unset and explicit `zero` both produce zero; only explicit `nan` produces NaN. `getZeroAttr` is correct for both integer and floating-point element types.
**Before:**
```
if (padding.has_value()) {
  TypedAttr padAttr = rewriter.getZeroAttr(memRefElementType);
  if (padding.value() == triton::PaddingOption::PAD_NAN) {
    ...
  }
  ...
  fillTensorWithOtherForMaskScenario(padVal, allocOp, boundarySizes, rewriter);
}
```
**After:**
```
TypedAttr padAttr = rewriter.getZeroAttr(memRefElementType);
if (padding.value_or(triton::PaddingOption::PAD_ZERO) ==
    triton::PaddingOption::PAD_NAN) {
  ...
}
...
fillTensorWithOtherForMaskScenario(padVal, allocOp, boundarySizes, rewriter);
```
The value_or(PAD_ZERO) idiom treats nullopt as PAD_ZERO, so only an explicit padding_option="nan" produces NaN; both "zero" and unset now produce zero.
### Why it's safe
- The mask path is reached only when mask != null — the no-mask early return at line 436 guarantees that. So unconditional fill is always meaningful here.
- rewriter.getZeroAttr(Type) returns the right IntegerAttr / FloatAttr for all element types (int, fp16, bf16, fp32, fp64).
- The downstream memref::SubViewOp + memref::CopyOp only overwrites the in-bounds subview, so the mask-out / out-of-bound lanes keep whatever was filled in.
### Performance impact
A microbenchmark was run on an Ascend NPU to measure the overhead introduced by the unconditional zero-fill on the masked-load path. The benchmark issues a tl.load(ptr, mask=m) over a contiguous 1-D tensor with BLOCK=4096, weeping three tensor sizes that span the L2 / HBM boundary. Numbers are wall-clock kernel time (microseconds), median over do_bench runs:
#### case1
| lens | grid |no fill (us) | fill zero (us) | Δ (us) | relative overhead |
|---|---|---|---|---|---|
| 65537 | (17,) | 2.125 | 2.321 | +0.196 | +9.22 % |
| 1048577 | (257,) | 5.447 | 5.893 | +0.446 | +8.19 % |
| 16777217 | (4097,) | 80.971 | 84.941 | +3.970 | +4.90 % |

```python
@triton.jit
def load_mask_other_zero(x_ptr, out_ptr, N, BLOCK: tl.constexpr):
    pid = tl.program_id(0)
    offs = pid * BLOCK + tl.arange(0, BLOCK)
    v = tl.load(x_ptr + offs, mask=offs < N, other=0.0)
    tl.store(out_ptr + offs, v, mask=offs < N)

@triton.jit
def load_without_other(x_ptr, out_ptr, N, BLOCK: tl.constexpr):
    pid = tl.program_id(0)
    offs = pid * BLOCK + tl.arange(0, BLOCK)
    # In this case scene, mask load without other will not fill zero
    v = tl.load(x_ptr + offs, mask=offs < N)
    tl.store(out_ptr + offs, v, mask=offs < N)
```
#### case2
| lens | grid |no fill (us) | fill zero (us) | Δ (us) | relative overhead |
|---|---|---|---|---|---|
| 65537 | (1,) | 3.813 | 3.964 | +0.151 | +3.96 % |
| 1048577 | (1,) | 45.229 | 45.428 | +0.199 | +0.44 % |
| 16777217 | (1,) | 1561.113 | 1563.951 | +2.838 | +0.18 % |

```python
@triton.jit
def load_mask_other_zero_loop(x_ptr, out_ptr, N, BLOCK: tl.constexpr):
    for idx in tl.range(0, N, BLOCK):
        offs = idx + tl.arange(0, BLOCK)
        v = tl.load(x_ptr + offs, mask=offs < N, other=0.0)
        tl.store(out_ptr + offs, v, mask=offs < N)

@triton.jit
def load_without_other_loop(x_ptr, out_ptr, N, BLOCK: tl.constexpr):
    for idx in tl.range(0, N, BLOCK):
        offs = idx + tl.arange(0, BLOCK)
        # In this case scene, mask load without other will not fill zero
        v = tl.load(x_ptr + offs, mask=offs < N)
        tl.store(out_ptr + offs, v, mask=offs < N)
```
**Observations:**
1. Relative overhead shrinks as len grows (9.22 % → 4.90 %). This is the expected behaviour: the copy itself is bandwidth-bound and its cost grows linearly with N, while the fill is a single linear write of the same size and therefore scales identically. The fixed part of the cost (kernel launch, subview, IR scheduling) is the same in both runs, so it dilutes the relative percentage at larger len.
2. **When doing it on separate cores, only one core handles fill zero while the others don't, making this an important factor. But when a single core does it in a loop, the impact of fill zero gets evened out**.

---

  - [x] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests

### DIFF
--- a/third_party/ascend/lib/TritonToLinalg/LoadStoreConverter.cpp
+++ b/third_party/ascend/lib/TritonToLinalg/LoadStoreConverter.cpp
@@ -401,21 +401,21 @@ LoadConverter::matchAndRewrite(triton::LoadOp op, OpAdaptor adaptor,
     } else {
       dstOffsets = srcOffsets;
     }
-    if (padding.has_value()) {
-      TypedAttr padAttr = rewriter.getZeroAttr(memRefElementType);
-      // triton already ensure only NAN and ZERO are passed in
-      if (padding.value() == triton::PaddingOption::PAD_NAN) {
-        // FIXME: Why NaN requires elemTy to be non-int or non-index?
-        assert(!memRefElementType.isIntOrIndex());
-        auto apNaN = llvm::APFloat::getNaN(
-            cast<FloatAttr>(padAttr).getValue().getSemantics());
-        padAttr = rewriter.getFloatAttr(memRefElementType, apNaN);
-      }
-      auto padVal = rewriter.create<arith::ConstantOp>(loc, padAttr);
-
-      fillTensorWithOtherForMaskScenario(padVal, allocOp, boundarySizes,
-                                         rewriter);
+    // When `padding` is not specified, default to zero. Both PAD_ZERO and
+    // unset (nullopt) map to 0; only PAD_NAN maps to NaN.
+    TypedAttr padAttr = rewriter.getZeroAttr(memRefElementType);
+    if (padding.value_or(triton::PaddingOption::PAD_ZERO) ==
+        triton::PaddingOption::PAD_NAN) {
+      // FIXME: Why NaN requires elemTy to be non-int or non-index?
+      assert(!memRefElementType.isIntOrIndex());
+      auto apNaN = llvm::APFloat::getNaN(
+          cast<FloatAttr>(padAttr).getValue().getSemantics());
+      padAttr = rewriter.getFloatAttr(memRefElementType, apNaN);
     }
+    auto padVal = rewriter.create<arith::ConstantOp>(loc, padAttr);
+
+    fillTensorWithOtherForMaskScenario(padVal, allocOp, boundarySizes,
+                                       rewriter);
     auto srcSubView = mlir::ConverterUtils::makeSubViewOp(
         ptr, srcOffsets, boundarySizes, loc, rewriter);
     auto dstSubview = mlir::ConverterUtils::makeSubViewOp(
@@ -477,16 +477,20 @@ LoadConverter::matchAndRewrite(triton::LoadOp op, OpAdaptor adaptor,
         op, "can not lower uncontinuout masked loads");
   }
 
+  // When mask is provided but no `other`, default the masked-out elements
+  // to zero.
+  Value effectiveOther;
   if (other) {
-    auto scalarOther =
-        mlir::ConverterUtils::getScalarValue(other, loc, rewriter);
-    assert(
-        scalarOther &&
-        "other value used in masked load produced by unsupported instruction!");
-
-    fillTensorWithOtherForMaskScenario(scalarOther, allocOp, mstate.dims,
-                                       rewriter);
+    effectiveOther = mlir::ConverterUtils::getScalarValue(other, loc, rewriter);
+    assert(effectiveOther &&
+           "other value used in masked load produced by unsupported "
+           "instruction!");
+  } else {
+    effectiveOther = rewriter.create<arith::ConstantOp>(
+        loc, rewriter.getZeroAttr(memRefElementType));
   }
+  fillTensorWithOtherForMaskScenario(effectiveOther, allocOp, mstate.dims,
+                                     rewriter);
 
   // To enable deinterleave optimization with mask load, mask state along last
   // dimension couldn't be split, which means `dims.back()` must be equal to
--- a/python/triton/language/semantic.py
+++ b/python/triton/language/semantic.py
@@ -1065,8 +1065,6 @@ class TritonSemantic(Generic[TensorTy]):
         # Check `boundary_check` argument
         boundary_check = self._canonicalize_boundary_check(boundary_check, dst_ty.get_block_shapes())
 
-        if boundary_check and padding is None:
-            padding = ir.PADDING_OPTION.PAD_ZERO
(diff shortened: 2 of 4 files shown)


third_party/ascend/lib/TritonToLinalg/LoadStoreConverter.cpp
```diff
@@ -401,21 +401,21 @@ LoadConverter::matchAndRewrite(triton::LoadOp op, OpAdaptor adaptor,
...
-        assert(!memRefElementType.isIntOrIndex());
-        auto apNaN = llvm::APFloat::getNaN(
-            cast<FloatAttr>(padAttr).getValue().getSemantics());
-        padAttr = rewriter.getFloatAttr(memRefElementType, apNaN);
-      }
-      auto padVal = rewriter.create<arith::ConstantOp>(loc, padAttr);
-
-      fillTensorWithOtherForMaskScenario(padVal, allocOp, boundarySizes,
-                                         rewriter);
+    // When `padding` is not specified, default to zero. Both PAD_ZERO and
```
hongziqi, line 404: It is recommended to add supplementary test cases, as there are no existing cases to guard against accuracy issues of this module.

third_party/ascend/lib/TritonToLinalg/LoadStoreConverter.cpp
```diff
@@ -401,21 +401,21 @@ LoadConverter::matchAndRewrite(triton::LoadOp op, OpAdaptor adaptor,
...
+    // unset (nullopt) map to 0; only PAD_NAN maps to NaN.
+    TypedAttr padAttr = rewriter.getZeroAttr(memRefElementType);
```
hongziqi, line 406: It is recommended to also remove the automatic zero-padding logic in the frontend semantic module.

WuTYSFG: Added it

third_party/ascend/lib/TritonToLinalg/LoadStoreConverter.cpp
KanuaK, line 404: Could we also apply the zero default to the scalar-loop paths used for unstructured memory accesses? Both TritonToUnstructure and BlockPtrAnalysis::rewriteAddPtrToUnstrucMemAcc initialize their loop-carried result with tensor.empty, rather than a zero-filled tensor.

WuTYSFG: I think it's neccessary, i will add it

PR #955, opened 2026-07-09
[TritonToUnstructure](fix) Fix bool masked store lowering in Unstructure

CHNJZ:
### Background

In the Ascend backend, a masked store on `bool` tensors may fail during the `TritonToLinalg` stage when both `cache_modifier` and `eviction_policy` are specified. A minimal trigger is:

```python
tl.store(
    y_ptr + offsets,
    value,
    mask=mask,
    cache_modifier=".cg",
    eviction_policy="evict_first",
)
```

where `x`, `y`, and `mask` are all `torch.bool` tensors.

For bool memory accesses, Triton may generate a `tt.bitcast` from `ptr<i1>` to `ptr<i8>`:

```mlir
%27 = tt.bitcast %arg1 : !tt.ptr<i1> -> !tt.ptr<i8>
ascend.indirect_store %27 : <i8>, ...
```

During `TritonToUnstructure`, the original logic uses the bitcasted `!tt.ptr<i8>` as the base pointer of `ascend.indirect_store`.

Later, in `TritonToLinalg`, `IndirectStoreConverter` expects `src` to have already been converted by the TypeConverter into a `MemRefType`. However, the actual `src` is still `!tt.ptr<i8>`, which causes the conversion to fail with `expected MemRefType for src`.

### Changes

This PR handles the special bool pointer bitcast before creating `ascend.indirect_store` in `TritonToUnstructure`:

```mlir
tt.bitcast ptr<i1> -> ptr<i8>
```

When this pattern is detected, the bitcasted `ptr<i8>` is no longer used as the base pointer of `ascend.indirect_store`. Instead, the original `ptr<i1>` is used.

This allows the later `TritonToLinalg` TypeConverter to consistently map `ptr<i1>` to `memref<?xi8>`, ensuring that `IndirectStoreConverter` can process the store correctly.

### Result

Before this fix:

```text
failed to legalize unresolved materialization from () to ('!tt.ptr<i8>')
see existing live user here: ascend.indirect_store ...
expected MemRefType for src
```

After this fix:

```mlir
ascend.indirect_store %arg1 : <i1>, ...
```

The subsequent type conversion correctly maps the bool pointer to `memref<?xi8>`, and the related bool masked store case can compile and execute successfully.

### Tests

A minimal regression test is added to cover the following scenario:

* bool input tensor
* bool dynamic mask
* `tl.store` with `cache_modifier`
* `tl.store` with `eviction_policy`

The test reliably reproduces the original issue and verifies that the fixed path compiles and runs successfully.

### DIFF
--- a/third_party/ascend/lib/TritonToUnstructure/UnstructureConversionPass.cpp
+++ b/third_party/ascend/lib/TritonToUnstructure/UnstructureConversionPass.cpp
@@ -191,6 +191,19 @@ LogicalResult tryRewriteIndirectFastPath(MemAccOpTy op, Location loc,
            "src must be ptr type");
     Value value = op.getValue();
     Value mask = op.getMask();
+
+    // For bool store, unwrap ptr<i1> -> ptr<i8> bitcast before creating
+    // indirect_store. Keep ptr<i1> so TypeConverter can map it to memref<?xi8>.
+    if (auto bitcastOp = srcPtr.getDefiningOp<triton::BitcastOp>()) {
+      auto srcPtrTy =
+          dyn_cast<triton::PointerType>(bitcastOp.getSrc().getType());
+      auto dstPtrTy = dyn_cast<triton::PointerType>(bitcastOp.getType());
+
+      if (srcPtrTy && dstPtrTy && srcPtrTy.getPointeeType().isInteger(1) &&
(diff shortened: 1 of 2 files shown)


hongziqi (approved):

KanuaK (approved):
lgtm

PR #999, opened 2026-07-13
[CI](fix) Fix doc preview display and resolve build failure

xuedinge233:
## Summary

Two improvements: RTD preview workflow runs independently, and the doc-build mock is made robust against future changes.

### 1. Decouple RTD Preview into a standalone workflow ([`docs-rtd-preview.yml`](.github/workflows/docs-rtd-preview.yml))

- **Trigger** changed from `workflow_run` (depending on a successful Documentation workflow) to `pull_request_target`, so fork PRs can also access the `RTD_TOKEN` secret
- **Path filters** limit triggering to doc-related changes: `docs/**`, `**/*.rst`, `**/*.md`, `**/requirements*.txt`, `**/setup.cfg`, `.readthedocs.yaml`
- **PR info** read directly from the `pull_request` payload instead of parsing `workflow_run.pull_requests`
- **documentation.yml** removes the `docs-rtd-preview.yml` paths entry; the two workflows are now fully decoupled

### 2. Doc-build mock robustness fixes ([`docs/zh/_mock/`](docs/zh/_mock/))

- **`_AutoModule`** — new auto-mocking module type that returns `MagicMock` for any undefined attribute, eliminating the need to manually add every new symbol
- **Fill in missing C-extension stubs**:
  - `triton._C.libtriton.native_specialize_impl`
  - `triton.backends.BaseBackend`
  - `triton._C.libtriton.gluon_ir` / `.linear_layout`
- **`_ascend_ir_stub.py`** — fix `SYNC_HINT` casing (C++ pybind11 uses lowercase `wait`/`set`/`internal`), add missing `EVENT` enum

## Why

Previously the RTD preview had to wait for the Documentation workflow to finish, and doc-unrelated PR changes still triggered the full pipeline. Now the preview responds directly to PR events — faster and cleaner.

The mock previously enumerated every C-extension symbol by hand; any new import added to the triton source would break the doc build. With `_AutoModule` as a safety net, undefined symbols are auto-mocked and only the few places that need real types (e.g. enums for `isinstance` checks) require precise stubs.


hipudding (dismissed):

.github/workflows/scripts/readthedocs_checkout.sh
```diff
@@ -33,14 +33,14 @@ echo "Building docs/${lang}/ for project ${READTHEDOCS_PROJECT} (${READTHEDOCS_L
 if [ "${READTHEDOCS_VERSION_TYPE:-}" = "external" ]; then
```
hipudding, line 33: I think this script should put in .github/workflows, because this file only used by this workflow

hipudding (dismissed):

hongziqi (approved):

hipudding (approved):

PR #1010, opened 2026-07-13
[compiler](feat) remove usage of enable-vf-fusion opt

ardazishvili:
vf-fusion going to be always enabled


ardazishvili:
Requested at https://gitcode.com/Ascend/AscendNPU-IR/pull/1342

PR #1019, opened 2026-07-14
[libdevice](fix) Libdevice operations support scalar and multi-dim te…

Levein7:
## Initiative
1. External Libdevice library operators cannot support scalar input.
2. External Libdevice library operators cannot support multi-dimensional tensor input.

## Solution
1. For scalar input/output, convert it to a 1-dimensional tensor for computation, then extract the scalar data.
2. For multi-dimensional tensor inputs, flatten them into a 1-dimensional tensor for computation, then reshape the result to the original shape.

## Additional Modify
1. Replace the "tensor.empty" relied on by CustomOp with "memref.alloc" to optimize the life-cycle management.
2. Switch the commit of the submodule "AscendNPU-IR" to 97f7c320, which ensures that the external Libdevice library operators are correctly inlined during compilation.

### DIFF
--- a/third_party/ascend/backend/utils.py
+++ b/third_party/ascend/backend/utils.py
@@ -586,11 +586,6 @@ def triton_support_ffts():
     return is_ffts_supported(arch) and (not force_disable_ffts())
 
 
-def triton_enable_libdevice_simt():
-    enable_libdevice_simt = os.getenv("TRITON_ENABLE_LIBDEVICE_SIMT", False)
-    return enable_libdevice_simt and is_compile_on_910_95()
-
-
 def get_cann_version_file_hash():
     ascend_path = _get_ascend_path()
     arch = get_machine_arch()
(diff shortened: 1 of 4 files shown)


third_party/ascend/backend/compiler.py
```diff
@@ -322,13 +322,24 @@ def bc_to_linalg_by_bishengir_opt(bc_data: bytes, metadata, opt):
...
+        cmd = [
             bishengir_opt_path,
             bc_path,
             "--mlir-print-debuginfo",
             "-o",
             mlir_path,
-        ], env=env, capture_output=True, check=True, text=True)
+        ]
+
+        if is_compile_on_910_95:
```
hongziqi: Both 910 and 950 use the bishengir-opt for 910, no need to add target

Levein7: Fixed

PR #1080, opened 2026-07-16
[user](feat) add plan_memory_strategy option

Shebaby55:
Add a option to change mem-plan order form defalut(declaration order) to largest-first(size order), which can reduce memory fragmentation and deal with UB overflow issue in certain cases. After changing the plan order, the set of buffers participating in reuse changes, which leads to changes of sync location and thus affect performance. In some cases, this may result in negative performance impacts. Therefore, we have added an option to control it. 

# New contributor declaration

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests

### DIFF
--- a/third_party/ascend/backend/compiler.py
+++ b/third_party/ascend/backend/compiler.py
@@ -700,6 +700,10 @@ def linalg_to_bin_enable_npu_compile_910_95(linalg: str, metadata, opt):
         if hfusion_enable_multiple_consumer_fusion:
             cmd_list += [f"--hfusion-enable-multiple-consumer-fusion={hfusion_enable_multiple_consumer_fusion}"]
 
+        plan_memory_strategy = metadata["plan_memory_strategy"]
+        if plan_memory_strategy is not None:
+            cmd_list += [f"--plan-memory-strategy={plan_memory_strategy}"]
+
         if opt.debug or os.getenv("TRITON_PRINT_AUTOTUNING", None) == "1":
             print(f"[DEBUG] cmd_list: {' '.join(cmd_list)}")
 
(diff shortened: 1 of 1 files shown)


third_party/ascend/backend/compiler.py
```diff
@@ -700,6 +700,10 @@ def linalg_to_bin_enable_npu_compile_910_95(linalg: str, metadata, opt):
         if hfusion_enable_multiple_consumer_fusion:
             cmd_list += [f"--hfusion-enable-multiple-consumer-fusion={hfusion_enable_multiple_consumer_fusion}"]
 
+        plan_memory_strategy = metadata["plan_memory_strategy"]
+        if plan_memory_strategy is not None:
+            cmd_list += [f"--plan-memory-strategy={plan_memory_strategy}"]
```
hongziqi, line 705: Is this compilation option used only for the 950?

third_party/ascend/backend/compiler.py
```diff
@@ -700,6 +700,10 @@ def linalg_to_bin_enable_npu_compile_910_95(linalg: str, metadata, opt):
         if hfusion_enable_multiple_consumer_fusion:
             cmd_list += [f"--hfusion-enable-multiple-consumer-fusion={hfusion_enable_multiple_consumer_fusion}"]
 
+        plan_memory_strategy = metadata["plan_memory_strategy"]
```
hongziqi, line 703: It is recommended that the description be supplemented to describe the background and the optimization effect.

third_party/ascend/backend/compiler.py
Shebaby55: yeah, only for 950, passing to 910 will cause compilation error

third_party/ascend/backend/compiler.py
Shebaby55: Should the option function be added here, or added to user documentation?

hongziqi: It can be added to the PR description first.

Shebaby55: added

hongziqi (approved):

KanuaK (approved):

KanuaK (commented):
lgtm

PR #1216, opened 2026-07-24
[compiler](fix) modify jit to fix specialization issue

KanuaK:
## Summary

Preserve value-based `constexpr` specialization when an integer kernel argument has an explicit Triton type annotation.

## Problem

`native_specialize_impl()` classifies a Python integer with value `1` as:

```python
("constexpr", 1)
```

For an argument annotated with an integer type such as `tl.int64`, `create_function_from_signature()` currently overwrites the first tuple element and produces:

```python
("i64", 1)
```

The value still participates in the specialization/cache key, but `_pack_args()` no longer recognizes the argument as a compiler constant. It therefore remains a dynamic kernel argument.

On the Ascend backend, this prevents downstream constant folding and range narrowing. In the original reproducer, the resulting dynamic index expression could trigger a UB overflow diagnostic with a large block size.

## Changes

- Preserve a `constexpr` classification already returned by the value specializer for explicitly annotated integer arguments.
- Apply the declared integer type only when the argument is not a specialized constant.
- Use actual booleans for `is_const`, `specialize`, and `align`, so Python control flow correctly honors `do_not_specialize`.-

| Input | Before | After |
| --- | --- | --- |
| `arg: tl.int64`, value `1` | `("i64", 1)` | `("constexpr", 1)` |
| `arg: tl.int64`, value `2` | `("i64", "")` | unchanged |
| `arg: tl.int64`, value `1`, `do_not_specialize=True` | `("i64", None)` | unchanged |

### DIFF
--- a/python/triton/runtime/jit.py
+++ b/python/triton/runtime/jit.py
@@ -398,14 +398,15 @@ def create_function_from_signature(sig, kparams, backend):
     assert len(sig.parameters) == len(kparams)
     # Create the function argument list and the dict entries for the return statement
     specialization = []
+    integer_annotation_types = {"i8", "i16", "i32", "i64", "u8", "u16", "u32", "u64"}
     # signature
     for name, kp in zip(sig.parameters.keys(), kparams):
         if kp.is_constexpr:
             specialization.append(f'("constexpr", {name})')
         else:
-            is_const = 'True' if kp.is_const else 'False'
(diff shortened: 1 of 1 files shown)


hongziqi (approved):

LH-123L (approved):
ok

WuTYSFG (approved):

PR #1242, opened 2026-07-27
[TritonToLinalg](fix) use absolute inputs for cos range reduction

### LINKED ISSUE #1202: [bug] [precision] `tl.floor(tl.cos(x))` returns `1` instead of `0` for small nonzero float32 inputs
### Describe the bug

On Ascend 910B3, a two-element Triton kernel computing `tl.floor(tl.cos(x))` produces a different result from the PyTorch reference for a small nonzero `float32` input. 

For the input value `-0.00034680962562561035`, the Ascend result is `1.0`, while `torch.floor(torch.cos(x))` returns `0.0`. The behavior is reproducible with a block size of 2. A block size of 1 does NOT reproduce it in the same environment, which may indicate that the issue is specific to a vector math lowering path.
```python
import torch
import torch_npu  # noqa: F401
import triton
import triton.language as tl


@triton.jit
def kernel(x_ptr, out_ptr, BLOCK: tl.constexpr):
    offsets = tl.arange(0, BLOCK)
    x = tl.load(x_ptr + offsets)
    tl.store(out_ptr + offsets, tl.floor(tl.cos(x)))


def main():
    x_cpu = torch.tensor(
        [-0.00034680962562561035, 0.0002011593896895647],
        dtype=torch.float32,
    )
    x = x_cpu.npu()
    actual = torch.empty_like(x)
    kernel[(1,)](x, actual, BLOCK=x.numel(), compile_mode="simd")
    torch.npu.synchronize()

    actual = actual.cpu()
    expected = torch.floor(torch.cos(x_cpu))
    print("input:    ", x_cpu.tolist())
    print("NPU:      ", actual.tolist())
    print("reference:", expected.tolist())
    assert not torch.equal(actual, expected)


if __name__ == "__main__":
    main()

```

output
```
input:     [-0.00034680962562561035, -0.00034680962562561035]
NPU:       [1.0, 1.0]
reference: [0.0, 0.0]
```

### Environment details

- Device: Ascend 910B3
- Architecture: AArch64
- CANN: 9.0.0
- Triton-Ascend commit: `fd24f3aad05c1c11f244be192c72c0b7ad7b96ae`
- Triton-Ascend package: `3.6.0+gitfd24f3aa`
- PyTorch: 2.7.1
- torch-npu: 2.7.1.post4
- Python: 3.11.15

shaoyuyoung:
Canonicalize tensor cos operands through abs so the even function uses the positive range-reduction path and avoids the one-ULP error around zero.

Add MLIR and runtime regressions for `floor(cos(x))`, including constant-folding convergence and fast-math preservation.

fixes #1202


shaoyuyoung:
hi, any update?
950 ci failure seems unrelated to this patch.

PR #1250, opened 2026-07-27
[CI](fix) Extend OCR timeout duration

xuedinge233:
### Summary

Some tasks may encounter ocr task timeout issues. Here, also set a timeout period in step to allow it to pass normally.

### DIFF
--- a/.github/workflows/ocr-review.yml
+++ b/.github/workflows/ocr-review.yml
@@ -70,7 +70,10 @@ permissions:
 jobs:
   code-review:
     runs-on: linux-amd64-cpu-8-hk
-    timeout-minutes: 30
+    timeout-minutes: 35
+    # Job-level timeout is a safety net for non-OCR steps (e.g. Get PR context).
+    # The OCR step uses its own step-level timeout with continue-on-error so a
+    # review timeout doesn't fail the check — Finalize step always exits 0.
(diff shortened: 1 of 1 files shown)


.github/workflows/ocr-review.yml
```diff
@@ -70,7 +70,8 @@ permissions:
 jobs:
   code-review:
     runs-on: linux-amd64-cpu-8-hk
-    timeout-minutes: 30
+    # No job-level timeout — OCR step uses its own step-level timeout
+    # with continue-on-error so a timeout doesn't fail the check.
```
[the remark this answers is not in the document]
xuedinge233: Modified.

.github/workflows/ocr-review.yml
```diff
@@ -70,7 +70,10 @@ permissions:
 jobs:
   code-review:
     runs-on: linux-amd64-cpu-8-hk
-    timeout-minutes: 30
+    timeout-minutes: 35
```
hongziqi, line 73: Do we need to sync this to main-dev?

hongziqi (approved):

.github/workflows/ocr-review.yml
xuedinge233: This is not needed. The current task is that all branches are effective.

WuTYSFG (approved):

PR #1346, opened 2026-08-03
[Interpreter](fix) TA 3.6 support interpreter

### LINKED ISSUE #1081: [RFC]: TA Interpreter For Precision Verifycation
### Motivation.

The Triton interpreter mode is enabled via the `TRITON_INTERPRET=1` environment variable, allowing the use of numpy as a backend to simulate the execution of Triton kernels without NPU/GPU dependency, primarily for debugging and precision testing.
The current operating process of the interpreter mode is shown in the following :

Triton Kernel
    │
    │ TRITON_INTERPRET=1
    ▼
InterpretedFunction (Entry Point)
    │
    ▼
Stage 1: AST Rewrite (first call, cached)
    · Wrap each assign RHS with to_tensor
    · Fix line numbers for debugging
    │
    ▼
Stage 2: GridExecutor
    │
    ├─► NPU/GPU → CPU Args Sync (shared storage ptr)
    │
    ├─► Patch Language Layer
    │     · builtin → builder wrapper
    │     · reduce / scan → numpy impl
    │
    ├─► Implicit Conversion
    │     int / Tensor / tuple → tl.tensor
    │
    ├─► Iterate Grid (x, y, z) serially
    │     │
    │     ▼
    │   Run Rewritten Kernel
    │     │
    │     ▼
    │   TritonSemantic (type infer / broadcast)
    │     │
    │     ▼
    │   InterpreterBuilder (numpy backend)
    │     · create_load / add / dot / atomic_rmw ...
    │     · returns TensorHandle(np.array, tl.dtype)
    │
    └─► CPU → NPU/GPU Result Writeback

The core idea is to simulate the API of the MLIR/LLVM IR Builder, but directly use NumPy to compute the result of each `create_*`.

### Proposed Change.

The overall design framework of Triton interpreter remains unchanged.
1. We  support all ops supported by Triton-lang in the TA interpreter mode.
2. We also support the interpreter mode implementation of  TA extensions  ops.

To minimize modifications to the community file interpreter.py, we introduce AscendInterpreterBuilder in interpreter.py ，the implementation of the  extension ops and the patch of tl.op are implemented in ascend_interpreter.py.

### Feedback Period.

_No response_

### CC List.

_No response_

### Any Other Things.

_No response_

zhuxuejie8:
## Background
The Ascend interpreter backend diverges from the compiler (and thus NPU hardware) in several semantics, so interpreter-mode results mismatch compiler-mode results for TA 3.6 kernels. This PR closes those gaps so the interpreter can be used as a reliable reference/debugging path for 3.6.

### 1. Match compiler numeric semantics
- fp16 reduction/scan : AscendReduceOps.sum and the new AscendScanOps.cumsum promote fp16 → fp32 before reducing and cast back, mirroring the compiler's optimization. Without this, interpreter produces different rounding from NPU for fp16 reductions. 
- Indexed reduce (argmin/argmax) : generic_reduce now treats a canonical (value, index) two-input combine specially, converting the index input to a per-axis local index so the result matches torch.min / torch.max (which return local, not global, indices).
- Integer division by zero : new create_idiv reproduces NPU saturation ( +//0 → dtype max , -//0 → dtype min , 0//0 → 0 ) instead of numpy's 0 , so saturating-div kernels behave identically.
- histogram : force int32 output — the base impl derives dtype from input weights and yields int64, which fails TensorHandle validation for the int32 return type.
- atomic MAX/MIN on float : new create_atomic_rmw bitcasts fp32/fp64 to int/uint of the same width and dispatches onto the existing integer atomics (community C++ only instantiates MAX/MIN over int types). 
### 2. Match compiler control-flow / indexing semantics via AST rewrite
Interpreter runs real Python, so and / or / not short-circuit and scalar[None, None] raises — both diverge from the compiler, which lowers these to element-wise & / | / ~ and expand_dims . _rewrite_boolops rewrites the kernel AST once (cached) before execution:

- and / or / not → & / | / ~
- subscripts whose slice contains None wrap their value in _ensure_indexable() (scalar → 0-d tensor) so broadcasting matches compiler mode.
### 3. Robust extension/builtin patching
- _GlobalsProxy lets _LangPatchScope patch directly from-imported builtins (e.g. from ...cann.extension import gather_out_to_ub ) — patching the source module attribute alone doesn't update the name bound in fn.__globals__ .
- _patch_lang_ascend switched from direct tl.x = ... assignment to scope.set_attr(...) , avoiding global pollution of tl across kernels.
### 4. vec_ops.extract_slice
sizes is now passed as list(sizes) to tl.block_type , bypassing an assertion that fires in interpreter mode; no effect on compile mode.
## Limits
-  extern_elementwise（including libdevice op,etc）、inline_asm、dot_scaled op are not supported, consistently with TA community 3.6
- bf16、fp8 dtype have precision problems in most cases, not repaired for now,  consistently with TA community 3.6

### DIFF
--- a/third_party/ascend/backend/ascend_interpreter.py
+++ b/third_party/ascend/backend/ascend_interpreter.py
@@ -31,8 +31,8 @@ import warnings
 import contextlib
 import numpy as np
 import triton.language as tl
-from triton.runtime.interpreter import InterpreterBuilder, TensorHandle, ReduceOps, _get_np_dtype, _patch_builtin, _LangPatchScope
-from triton._C.libtriton import interpreter as _interpreter
+from triton.runtime.interpreter import InterpreterBuilder, TensorHandle, ReduceOps, ScanOps, _get_np_dtype, _patch_builtin, _patch_attr, _LangPatchScope
+from triton._C.libtriton import interpreter as _interpreter, ir as _ir
 
 
 class AscendReduceOps(ReduceOps):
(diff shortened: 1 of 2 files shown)


third_party/ascend/backend/ascend_interpreter.py
```diff
@@ -31,8 +31,8 @@
 import contextlib
 import numpy as np
 import triton.language as tl
-from triton.runtime.interpreter import InterpreterBuilder, TensorHandle, ReduceOps, _get_np_dtype, _patch_builtin, _LangPatchScope
-from triton._C.libtriton import interpreter as _interpreter
+from triton.runtime.interpreter import InterpreterBuilder, TensorHandle, ReduceOps, ScanOps, _get_np_dtype, _patch_builtin, _patch_attr, _LangPatchScope
```
hongziqi, line 34: We suggest associating this issue with the following link: https://github.com/triton-lang/triton-ascend/issues/1081

hongziqi (approved):

WuTYSFG (approved):

PR #1352, opened 2026-08-03
[dot](feat) add al.dot with per-operand fractal (zN) layout

### LINKED ISSUE #1354: [Feature]: Supports extension.dot op
# [dot] Introduce `al.dot` — per-operand fractal (zN) matmul for the nz2nz cube path

## 1. Support Matrix for `al.dot` (Ascend 950 Series)

**Prerequisite:** a fractal operand must be 4D with a static shape; a non-fractal operand must be 2D.

### 1.1 Per-operand format combinations

`format_a` / `format_b` / `format_c` are independent. All 8 combinations are supported:

| `format_a` | `format_b` | `format_c` | Meaning | Status |
|---|---|---|---|---|
| fractal | fractal | fractal | fully nz — cube in/out stays fractal | ✔️ |
| fractal | fractal | nd | nz inputs, ND result (ND store / ND consumer) | ✔️ |
| fractal | nd | fractal | mixed operand, fractal result | ✔️ |
| fractal | nd | nd | mixed operand, ND result | ✔️ |
| nd | fractal | fractal | mixed operand, fractal result | ✔️ |
| nd | fractal | nd | mixed operand, ND result | ✔️ |
| nd | nd | fractal | plain ND inputs, fractal result | ✔️ |
| nd | nd | nd | equivalent to a plain 2D matmul | ✔️ |

### 1.2 dtype

Output dtype is the **cube L0C accumulator dtype**, independent of the input dtype.

| input dtype | elem bytes | zN block `[16, b]`, `b = 32 / elem_bytes` | output dtype | `format_c="fractal"` block |
|---|---|---|---|---|
| float16 | 2 | `[16, 16]` | float32 | `[16, 16]` |
| bfloat16 | 2 | `[16, 16]` | float32 | `[16, 16]` |
| float32 | 4 | `[16, 8]` | float32 | `[16, 16]` |
| int8 | 1 | `[16, 32]` (lhs) / `[32, 32]` (rhs) | int32 | `[16, 16]` |
| float64 / float8 / float4 / int16 | — | ❌ | ❌ | ❌ |

`A` and `B` must have the same element dtype.

### 1.3 Limitations & fallback rules

1. `format_c="fractal"` requires `M % 16 == 0` and `N % 16 == 0`.
2. **int8 right operand must be loaded as `[32, 32]`** on Ascend 950 — the cube fractal row block is
   per-operand.
3. No `hf32` selection for float32 inputs — the emitted `linalg.matmul` is fixed to
   `input_precision = "ieee"`.

---

## 2. Background & Motivation

On Ascend, cube (matmul) operands live in a fractal (NZ) layout while the vector unit and global
memory work in ND. Today every cube boundary therefore inserts a layout conversion:

- GM → cube: `nd2nz`
- cube → GM / cube → vector: `nz2nd`

For cube-heavy and cube-vector (CV) fused kernels these conversions dominate: a CVC chain pays
`nd2nz → mmad → nz2nd → vector → nd2nz → mmad → nz2nd`, even though every intermediate could have
stayed fractal. The conversions cost both DMA bandwidth and UB/L1 footprint, and in CV kernels they
force the cube result out to a GM workspace instead of staying on-chip.

`tl.dot` cannot express this: it takes ND operands and produces an ND result, so the frontend has no
way to tell the backend "this tensor is already fractal, do not convert it".

**nz2nz** means the cube's inputs and outputs enter and leave in fractal form, so the cube↔GM and
cube↔vector boundaries carry no `nd2nz`/`nz2nd`. To let the backend fold those boundaries, the
frontend first needs a way to *declare* per-operand layout. That is what `al.dot` adds.

---

## 3. Scope

- **New op:** `al.dot` (MLIR `ascend.dot`)
- **Target scenarios:** pure-cube nz chains, CV epilogue, VC prologue, CVC, fan-in, and mixed
  fractal/ND operands
- **Target device:** Ascend 950 series (910_9589 / A5)
- **Target pass:** `TritonToLinalg` (new `DotConverter`)
- **Non-goal:** no new backend op. The frontend decomposes into ops the backend already consumes;
  the actual nz2nz optimisation is done by existing backend layout inference and convert folding.

### 3.1 Full workflow

```
user
  al.dot(a, b, format_a=..., format_b=..., format_c=...)          [strings]
    │
    ├─ core.py:dot            @builtin
    │    ├─ _dot_operand_is_fractal   string -> bool (parsed only here)
    │    ├─ _check_dot_fractal_block  validate trailing [block_row, block_col]
    │    ├─ _dot_to_nd_shape          zN -> ND
    │    └─ output_shape              [N/16, M/16, 16, 16] if fractal_c else [M, N]
    │
    ├─ semantic.py:dot        emit (bool)
    │    └─ _semantic.builder._ascend_builder.create_dot(...)
    │            NOTE: goes through _ascend_builder explicitly, NOT the
    │            setup_unified_builder allow-list — `create_dot` collides with
    │            upstream ir.builder.create_dot used by tl.dot.
    │
    ├─ ascend_ir.cc:create_dot        pybind glue on AscendNPUIROpBuilder
    │
    └─ ascend.dot                     attrs fractal_a / fractal_b / fractal_c : BoolAttr
         ├─ DotOp::inferReturnTypes   type contract, checked at ast_to_ttir time
         ├─ DotOp::verify             rank <-> fractal flag
         └─ DotConverter              TritonToLinalg
               │
               ├─ fractal input  ->  hivm.hir.convert_layout {Fractal -> ND}
               ├─ K reconciliation (zero-pad the short ND side)
               ├─ linalg.matmul, mixed precision (f16 in -> f32 acc)
               └─ fractal output ->  hivm.hir.convert_layout {ND -> Fractal[16,16]}
                     │
                     └─ backend (AscendNPU-IR): InferHIVMDataLayout,
                        PropagateConvertLayout, CombineOptimizedConvertLayout,
                        NormalizeMatmul, InlineFixpipe
                        -> the boundary converts fold away, mmad reads the
                           fractal buffer directly, fixpipe emits nz2nz
```

---

### 4.1 API boundary

`format_*` are strings only at the user entry point (`core.py:dot`); `semantic` → pybind → op
attribute all use bool.

### 4.2 zN convention

Non-transposed `A` and `B` are **both zN**, so K lands at the same relative position in both and the
fractal→ND formula is identical:

| operand | ND | zN 4D `[d0, d1, 16, b]` | Fractal → ND |
|---|---|---|---|
| A (lhs, input dtype) | `[M, K]` | `[K/16, M/16, 16, b]` | `[d1·d2, d0·d3]` = `[M, K]` |
| B (rhs, input dtype) | `[K, N]` | `[N/b, K/16, 16, b]` | `[d1·d2, d0·d3]` = `[K, N]` |
| C (result, **acc dtype**) | `[M, N]` | `[N/16, M/16, 16, 16]` | — |

> B's outer dims are `[N1, K1]`, **not** `[K1, N1]`. `[K1, N1]` is nZ, and the cube reads a
> non-transposed rank-4 B as zN — so nZ silently swaps K and N instead of failing to compile.

The emitted `convert_layout` carries the generic `Fractal` layout; the concrete nZ/zN orientation is
decided by the backend from `DOTA_ND` / `DOTB_ND` / `DOTC_ND` plus the transpose flag. There is no
transpose flag on `al.dot` itself — transposition is written as `tl.trans` / `tl.permute` before the
call and folded by the backend.

### 4.3 Accumulator dtype

Float inputs accumulate in f32, int8 in i32, so the result carries the accumulator dtype regardless
of input dtype, and the `format_c="fractal"` block is always `[16, 16]` (the L0C fractal, not
`[16, b]`) — making the fixpipe a straight move with no downcast and no re-tile. `DotConverter` keeps
`inElemTy` separate from `accTy` so the `linalg.matmul` is genuinely mixed-precision.

### 4.4 K reconciliation

Only when exactly one operand is fractal: the fractal side's K is block-rounded while the ND side may
be shorter, so the short side is zero-padded to match (safe — the extra K products are ×0). Padding
is allowed only when the gap is under one block; a larger mismatch is a compile error. When both
operands are fractal, K is guaranteed consistent by the two 4D shapes and nothing is done.

`zeroPadDimTo` builds the pad from `memref.alloc` + `linalg.fill(0)` + `subview` + `copy` rather than
`tensor.insert_slice`, which lowers to a scalar copy loop that blows up cbuf.

### 4.5 Cube kernel tagging

`ascend.dot` is counted into `existDotFlag`; otherwise a kernel whose only matmul is an `al.dot` gets
tagged `mix_mode = "aiv"` and the cube tile-and-slice overflows cbuf.

---

## 5. Implementation Components

| layer | file | responsibility |
|---|---|---|
| op definition | `include/Dialect/TritonAscend/IR/TritonAscendOps.td` | `DotOp`: `Pure` + `InferTypeOpInterface`, attrs `fractal_a/b/c`, `hasVerifier` |
| type / verify | `lib/Dialect/TritonAscend/IR/TritonAscendOps.cpp` | `DotOp::inferReturnTypes` (zN conversion + accumulator dtype + fractal_c shape), `DotOp::verify` (rank ↔ fractal) |
| shared constants | `include/Utils/Utils.h` | `kFractalBlock`, `kBytesPerFractalCol`, `kBitsPerByte`, `kDotAccIntWidth` |
| pybind | `ascend_ir.cc` | `create_dot` glue on `AscendNPUIROpBuilder` |
| builtin | `language/cann/extension/core.py` | `dot`, `_dot_operand_is_fractal`, `_dot_to_nd_shape`, `_check_dot_fractal_block` |
| emit | `language/cann/extension/semantic.py` | emit via `_ascend_builder`, attach accumulator-dtype `block_type` |
| lowering | `lib/TritonToLinalg/TritonOpConverter.{h,cpp}` | `DotConverter`, `dotFractalToND`, `reconcileDotContractionK`, `zeroPadDimTo`, `findRedirectableLoad` |
| pass registration | `lib/TritonToLinalg/TritonToLinalgPass.cpp` | register `DotConverter`; count `ascend.dot` into `existDotFlag` |
| docs | `docs/zh/python-api/*` | constraints entry, example, API index |

**Invariant:** the zN conversion exists in three places — `DotOp::inferReturnTypes` (C++ contract),
`_dot_to_nd_shape` (Python), `dotFractalToND` (lowering). All three must change together or
`ast_to_ttir` type checking fails.

---

## 6. Validation status

Test design and cases are tracked under backend SR **SR20260713312453**
(【NPUIR】【A5】【CV】MMADL1支持NZ格式) in AscendNPU-IR-DT:
https://gitcode.com/NPU-IR/AscendNPU-IR-DT/pull/53

- **UT** — 30 lit cases over `normalize-matmul.mlir`, `infer-hivm-data-layout.mlir`,
  `combine-optimized-convert-layout.mlir`, `convert_layout.mlir`, `inline-fixpipe.mlir`,
  `matmul.mlir`.
- **ST** — 28 end-to-end cases in `src/modules/cube_vector/test_dot_nz2nz.py`, each compared against
  a torch reference with `assert_close`.

https://github.com/triton-lang/triton-ascend/pull/1352

zengzengran:
Added Ascend extended operator `al.dot` (MLIR `ascend.dot`) supporting matrix multiplication `D = A · B` by declaring operand types (zN)

https://github.com/triton-lang/triton-ascend/issues/1354

# New contributor declaration

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests

### DIFF
--- a/third_party/ascend/language/cann/extension/core.py
+++ b/third_party/ascend/language/cann/extension/core.py
@@ -593,24 +593,20 @@ def _check_dot_fractal_block(name, shape, is_lhs, dtype):
     block_col = _unwrap_if_constexpr(shape[-1])
 
     if block_col != expected_col:
-        raise ValueError(
-            f"dot nZ operand `{name}`: block col dim is {block_col}, expected "
-            f"{expected_col} (= 32 / {elem_bytes} bytes per {dtype} element)")
+        raise ValueError(f"dot nZ operand `{name}`: block col dim is {block_col}, expected "
+                         f"{expected_col} (= 32 / {elem_bytes} bytes per {dtype} element)")
 
     if elem_bytes != 1:
         if block_row != 16:
-            raise ValueError(
-                f"dot nZ operand `{name}`: block row dim is {block_row}, expected 16")
+            raise ValueError(f"dot nZ operand `{name}`: block row dim is {block_row}, expected 16")
         return
 
     if block_row not in (16, 32):
-        raise ValueError(
-            f"dot int8 nZ operand `{name}`: block row dim is {block_row}, expected 16 or 32")
+        raise ValueError(f"dot int8 nZ operand `{name}`: block row dim is {block_row}, expected 16 or 32")
     if not is_lhs and block_row == 16:
-        raise ValueError(
-            f"dot int8 right operand `{name}`: block row dim is 16, but a non-transpose "
-            f"`{name}` needs 32 (block [32,32], i.e. zN[N/32, K/32, 32, 32]); a [16,32] "
-            f"block forces an unsupported layout conversion in the backend")
+        raise ValueError(f"dot int8 right operand `{name}`: block row dim is 16, but a non-transpose "
+                         f"`{name}` needs 32 (block [32,32], i.e. zN[N/32, K/32, 32, 32]); a [16,32] "
+                         f"block forces an unsupported layout conversion in the backend")
 
 
 def _dot_operand_is_fractal(name, fmt):
@@ -625,15 +621,12 @@ def _dot_operand_is_fractal(name, fmt):
         return False
     if fmt == "fractal":
         return True
-    raise ValueError(
-        f"dot `{name}`: format must be 'fractal', 'nd', or '' "
-        f"(empty/unset = non-fractal ND); got {fmt!r}")
+    raise ValueError(f"dot `{name}`: format must be 'fractal', 'nd', or '' "
+                     f"(empty/unset = non-fractal ND); got {fmt!r}")
 
 
 @builtin
-def dot(a: tl.tensor, b: tl.tensor,
-        format_a="", format_b="", format_c="",
-        _semantic=None) -> tl.tensor:
+def dot(a: tl.tensor, b: tl.tensor, format_a="", format_b="", format_c="", _semantic=None) -> tl.tensor:
     """
     Matrix multiply ``D = A * B`` with per-operand layout format.
 
--- a/third_party/ascend/language/cann/extension/core.py
+++ b/third_party/ascend/language/cann/extension/core.py
@@ -24,7 +24,7 @@ __all__ = [
     "ascend_address_space", "builtin", "CORE", "copy_from_ub_to_l1", "copy", "debug_barrier", "fixpipe",
     "FixpipeDMAMode", "FixpipeDualDstMode", "FixpipePreQuantMode", "FixpipePreReluMode", "int64", "is_builtin", "MODE",
     "PIPE", "SYNC_HINT", "EVENT_ID", "IteratorType", "sub_vec_id", "sub_vec_num", "sync_block_all", "sync_block_set",
-    "sync_block_wait", "SYNC_IN_VF", "conv1d"
+    "sync_block_wait", "SYNC_IN_VF", "conv1d", "dot"
 ]
 
 import enum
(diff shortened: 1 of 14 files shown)


third_party/ascend/language/cann/extension/core.py
```diff
@@ -567,3 +567,111 @@ def _check_and_normalize_1d_param(param, name):
...
+    if fmt is None or fmt == "" or fmt == "nd":
+        return False
+    if fmt == "fractal":
+        return True
+    raise ValueError(f"dot `{name}`: format must be 'fractal', 'nd', or '' "
+                     f"(empty/unset = non-fractal ND); got {fmt!r}")
+
+
+@builtin
+def dot(a: tl.tensor, b: tl.tensor, format_a="", format_b="", format_c="", _semantic=None) -> tl.tensor:
```
hongziqi, line 629: It is recommended to add end-to-end test cases under `third_party/ascend/pytest_ut`.

winter77-x (dismissed):

hongziqi (dismissed):

third_party/ascend/language/cann/extension/core.py
zengzengran: IR test has been added. triton-ascend\third_party\ascend\unittest\Conversion\General\TritonToLinalg\dot_fractal.mlir
The latest TA and A5 AscendNPU IR are required for the end-to-end test.

winter77-x (approved):

hongziqi (approved):

PR #1357, opened 2026-08-03
[costmodel](feat) use costmodel in TA autotuner

BloodOrange666:
## Motivation

Triton-Ascend already provides `costmodel_bench` for predicting the performance of generated TTIR, but it was not integrated into Triton-Ascend's autotuning flow. As a result, autotune still had to compile and benchmark every candidate configuration on real hardware.

This PR integrates the Ascend Costmodel as an optional autotune prefilter. It reduces the number of configurations entering real NPU benchmarking, while the final `best_config` is still selected using hardware measurements.

## Changes

- Add Costmodel pruning through `prune_configs_by["costmodel"]`:

  ```python
  prune_configs_by={
      "costmodel": {
          "top_k": 0.25
      },
  }
  ```

- Preserve the existing pruning order:

  ```text
  candidate configs
    -> early_config_prune
    -> costmodel or perf_model
    -> real hardware benchmark
    -> best_config
  ```

- Make `costmodel` and `perf_model` mutually exclusive.
- Generate TTIR and runtime argument bindings automatically for each candidate.
- Keep Costmodel-equivalent configurations as a group to avoid pruning differences that are not visible to the model.
- Fall back to the configurations after early pruning when Costmodel generation or prediction fails.
- Benchmark lower-ranked fallback configurations when all primary configurations fail to compile or run.
- Validate `top_k` strictly.
- Detect the current SoC through `rtGetSocVersion()` and map supported SoCs to the corresponding hardware profile:
  - Ascend 910B and 910_93 variants use `ascend_910b.json`;
  - Ascend 910_95, Ascend 950, and David variants use `ascend_davidv100.json`.
- Package the default hardware JSON profiles in the wheel.
- Support separate and legacy combined `arg-bindings` / `hardware-config` option forms in the native in-process Costmodel interface.
- Keep the existing direct `costmodel_bench` interface compatible with the Inductor integration.
- Add English and Chinese usage documentation and examples.

## Failure Handling

Costmodel is only used for pruning and does not affect correctness:

- A failure for one candidate does not prevent other candidates from being predicted.
- If no candidate receives a valid prediction, autotune falls back to real benchmarking.
- Unknown SoCs do not silently use an incompatible hardware profile.
- Real NPU measurements still determine the final configuration.


python/triton/runtime/autotuner.py
```diff
@@ -259,11 +282,40 @@ def benchmark():
...
                     "No valid autotuner configs after pruning. `early_config_prune` should return at least one config.")
+        if self.costmodel_options is not None:
+            try:
+                from triton.compiler.compiler import make_backend
+
+                backend = make_backend(driver.active.get_current_target())
+                hook = getattr(backend, "prune_configs_by_costmodel", None)
+                if hook is None:
+                    warnings.warn(
+                        f"Backend {type(backend).__name__} does not support costmodel pruning; "
```
zhanwei33: Which backends currently support costmodel tuning?

zhanwei33:
## Review findings

### [P1] Package the default hardware profiles with the wheel

`_resolve_hardware_config()` automatically looks up `ascend_910b.json` / `ascend_davidv100.json` when `hardware_config` is not explicitly supplied. In the source checkout the second candidate resolves to `third_party/ascend/costmodel/configs/`, but the wheel installs only `third_party/ascend/backend` as `triton/backends/ascend`. This PR does not add a copy or package-data rule for the two JSON profiles.

I checked a recent built wheel: it contains `triton/backends/ascend/runtime/costmodel_runtime.py`, but neither profile JSON. Therefore the documented default:

```python
prune_configs_by={"costmodel": True}
```

will raise `FileNotFoundError` while resolving the automatic profile, then fail open in the generic autotuner and benchmark all original configs. This preserves correctness, but makes Costmodel pruning ineffective for wheel users.

Please package the profiles under a stable package path (for example `triton/backends/ascend/costmodel/configs/`), resolve them as package resources, and add an installed-wheel test covering default 910B and A5 profile resolution.

### [P2] Validate `min_configs` explicitly

`top_k` is type/range checked, while `min_configs` is passed directly to `int(min_configs)`. Thus `1.9` is silently truncated, and `True` or `"2"` are accepted. Please require a non-bool positive integer at the options boundary and add parameterized tests for zero, negative, float, bool, and string inputs.

The generic hook/fallback flow and the new separate C++ argv parsing look consistent. It would still be useful to add native/inproc coverage for separate argv, comma-combined, and space-combined `arg-bindings` / `hardware-config` forms.

third_party/ascend/backend/runtime/costmodel_runtime.py
```diff
@@ -29,6 +33,147 @@ def _costmodel_cache_namespace() -> str:
...
+    if arch.startswith(("ascend910b", "ascend91093")):
+        return _ASCEND_910B_CONFIG
+
+    raise ValueError(f"Costmodel has no hardware profile for target arch {target_arch!r}")
+
+
+def _costmodel_config_candidates(filename: str):
+    runtime_dir = os.path.dirname(__file__)
+    return [
+        os.path.join(runtime_dir, "../../../costmodel/configs", filename),
```
zhanwei33, line 74: [P1] These candidates work from the source tree, but not from the installed wheel. The wheel installs only `third_party/ascend/backend` as `triton/backends/ascend`; it does not contain either JSON profile under any of the three candidate roots. Default `prune_configs_by={"costmodel": True}` therefore raises `FileNotFoundError` during profile resolution and fails open to benchmarking every config. Please package the profiles under a stable package path and add an installed-wheel default-profile test.

third_party/ascend/triton_ascend.cc
```diff
@@ -262,6 +266,14 @@ parseAscendCostmodelArgs(const std::vector<std::string> &extraArgs) {
       }
       continue;
     }
+    // Prefer one top-level option per argv entry. This avoids overloading the
+    // comma used inside arg-bindings as a pass-option separator. Keep the
+    // combined option-string parser above for backwards compatibility.
+    if (arg.rfind("arg-bindings=", 0) == 0 ||
```
zhanwei33, line 311: [P2][test] This new branch is required for the Python side to pass `arg-bindings` and `hardware-config` as separate argv entries, but the added Python tests mock the C API and do not exercise this parser. Please add a native/inproc test covering separate argv plus the legacy comma- and space-combined forms, so a future parser change cannot silently drop the hardware profile.

python/triton/runtime/autotuner.py
BloodOrange666: Only Ascend backend is support, so this warning message is unnecessary and will be deleted.

third_party/ascend/backend/runtime/costmodel_runtime.py
BloodOrange666: Yes, the current source-tree fallback paths do not cover the installed-wheel layout. I will package the three hardware configuration files under `triton/backends/ascend/costmodel/configs`, resolve the default profile from that stable package-relative path, and add an installed-wheel test covering both the 910B and A5 profile mappings.

third_party/ascend/triton_ascend.cc
BloodOrange666: I will add an in-process integration test using the real `run_costmodel_inproc` binding. It will cover separate argv entries and the legacy comma- and space-combined forms, verify equal predictions, and check that an invalid hardware path is not silently ignored.

third_party/ascend/patch/triton-ascend-dev-3.6.0.patch
```diff
@@ -1,22 +1,127 @@
...
+
+         self.fn = fn
+         self.base_fn = fn
+@@ -126,7 +143,7 @@ class Autotuner(KernelInterface):
          return self._do_bench
- 
+
      def _bench(self, *args, config, **meta):
 -        from ..compiler.errors import CompileTimeAssertionFailure
 +        from ..compiler.errors import CompileTimeAssertionFailure, MLIRCompilationError
```
LH-123L: dev补丁只在开发态生效  release版本不带，业务需要，则需要迁移到triton-ascend-3.6.0.patch

third_party/ascend/patch/triton-ascend-dev-3.6.0.patch
```diff
@@ -1,22 +1,127 @@
...
          try:
              return self.do_bench(kernel_call, quantiles=(0.5, 0.2, 0.8))
 -        except (OutOfResources, CompileTimeAssertionFailure, PTXASError) as e:
 +        except (OutOfResources, CompileTimeAssertionFailure, MLIRCompilationError) as e:
              if verbose:
                  print(f"Autotuning failed with {e}")
              return [float("inf"), float("inf"), float("inf")]
+@@ -227,6 +244,12 @@ class Autotuner(KernelInterface):
+                 def benchmark():
+                     bench_start = time.time()
```
LH-123L: third_party\ascend\backend\runtime\autotuner.py  ascend的autotune实现均在这个里面 看下是否可以迁移到third_party下面  从而避免侵入式修改

PR #1397, opened 2026-08-05
[ssbuffer](fix) fix multi output in multi-buffer innerscope

cxtverygood:
This pull request is designed to fix the scenario where one op has multiple outputs which are tensor-type dependent values.


third_party/ascend/lib/DynamicCVPipeline/AllocMultiCache/AddMultiBufferInnerScope.cpp
```diff
@@ -507,14 +507,15 @@ static bool isEmptyFillPattern(Value depVal) {
 SmallVector<Value>
```
cxtverygood, line 507: 补下UT

Skyminers (approved):

PR #1417, opened 2026-08-06
[DeInterLeave](fix) Fix nonzero-base deinterleave lowering

zhanwei33:
## Problem

The affected FlagGems test is:

```bash
tests/test_fused_indexer_q_rope_quant.py::test_fused_indexer_q_rope_quant
```

The TA compile failure is in the FP8 path (`use_fp4=False`) for both `cache_dtype=torch.float32` and `cache_dtype=torch.bfloat16`. The relevant Triton kernel is `_fused_indexer_q_rope_quant_kernel`.

Its GPT-J RoPE access has a nonzero NoPE base followed by the canonical even/odd stride-2 loads:

```python
rot_base = base_ptr + INDEX_Q_NOPE_DIM  # 64 for this test
x_even = tl.load(rot_base + half_offset * 2)
x_odd = tl.load(rot_base + half_offset * 2 + 1)
```

This is valid pointer arithmetic: `+64` belongs to the base address, while only the final `+0` / `+1` selects the even or odd lane. The old `InterleaveOptimization` recursively treated every constant `arith.addi` in the offset expression as a lane marker and asserted unless it was `+1`. Consequently the compiler stopped in `TritonToLinalg` with:

```text
Arith::constant value of addi's operand must be 1 when
calculate deinterleave offset
```

After this blocker is removed, the same kernel also reaches an independent scalar `tt.precise_divf` conversion issue: `PreciseDivConverter` casts the original result to `RankedTensorType`, which is null for a scalar result.

## Fix

- Normalize additive offsets into a preserved base expression and an even/odd lane mode. Constants inside the base are no longer interpreted as lane selectors.
- Preserve the normalized static offset in the generated `memref.reinterpret_cast`, and make unsupported/unprovable expressions return `failure()` rather than assert.
- Apply the same normalization to unmasked/masked deinterleave loads and unmasked/masked interleave stores.
- Lower `tt.precise_divf` from converted operands and let `arith.divf` infer the scalar or tensor result type.
- Add nonzero-base regression coverage: unmasked `bias=64` and `bias=65`, plus masked load/store coverage at `bias=64`.

### `PreciseDivConverter` rationale

This change fixes a second, independent blocker exposed after the deinterleave assertion is removed. `tt.precise_divf` is valid for scalar as well as tensor floating-point values. The previous `dyn_cast<RankedTensorType>(op.getResult().getType())` is a checked downcast, not a numerical type conversion: it returns a null type for a scalar result. Passing that null type to the explicit-result-type `arith.divf` builder caused the scalar conversion failure.

For valid TTIR, `tt.precise_divf` has the `SameOperandsAndResultType` constraint: its two operands and result have the same complete MLIR type. Therefore `arith.divf` can infer its scalar or tensor result type directly from the converted operands. If an explicit result type were needed, the correct equivalent would be `opa.getType()`, not a `RankedTensorType` downcast of the original result.

The converter now uses `adaptor.getX()` and `adaptor.getY()`, which are the operands after dialect conversion. This is conversion correctness and makes the target `arith.divf` consume legal target-side values; it is not the direct cause of the scalar crash. Replacing the op with `divOp.getResult()` then maps the single original result to the inferred target result.

The added equality check compares the complete converted MLIR types, not only the scalar dtype. It is defensive: it should not fail for valid TTIR. It does not insert a numerical cast. A mismatch means either malformed/unverified TTIR or an inconsistent TypeConverter mapping, and must be fixed before this lowering rather than by arbitrarily casting one operand here.

Numerical promotion belongs to the DSL/TTIR construction stage. The DSL first legalizes binary operands, emitting explicit cast IR when required (or directly materializing a scalar constant in the common type), then creates same-typed `tt.precise_divf`. In particular, `tl.math.div_rn` accepts FP32 tensor operands; FP16/BF16 tensor inputs must be explicitly cast to FP32 before calling it. `PreciseDivConverter` only lowers that already-legal exact division to `arith.divf`.

### 1. FlagGems device test

On a pre-fix TA wheel with the Ascend runtime environment loaded, the smallest FP8/FP32-cache parameterized case is:

```bash
source /home/w00609825/set_env.sh
cd /home/w00609825/triton_workspace/FlagGems_new
conda run --no-capture-output -n 0802_ta36 \
  python -m pytest -q \
  'tests/test_fused_indexer_q_rope_quant.py::test_fused_indexer_q_rope_quant[False-cache_dtype0-1]'
```

`cache_dtype0` is `torch.float32`; `cache_dtype1` is `torch.bfloat16`. Both `use_fp4=False` cache signatures exercise the same failing TA compile path. This command is a device functional test, so it initializes tensors and may execute a kernel.

### 2. Offline A5 compiler reproducer

The compile-only reproducer does not obtain a launcher, load a binary, or execute an NPU kernel. It explicitly compiles for `GPUTarget("npu", "Ascend910_9589", 0)` with `use_bytecode=True` and covers both cache signatures:

```bash
source /home/w00609825/set_env.sh
conda run --no-capture-output -n 0802_ta36 \
  python /home/w00609825/triton_workspace/.tasks/flaggems算子TA运行问题/test_fused_indexer_q_rope_quant_a5_compile.py
```

The pre-fix wheel fails in `TritonToLinalg` with the assertion above. With a wheel built from this PR, run the same command in that wheel's environment; both `cache_fp32` and `cache_bf16` should succeed and emit `ttadapter`, `mlirbc`, `bcmlir`, and `npubin`.

### 3. Triton-Ascend regression test

In a checkout built from this PR:

```bash
cd /path/to/triton-ascend
python -m pytest -q \
  third_party/ascend/unittest/pytest_ut/test_interleave_optimizaiton.py
```

Recorded result: `7 passed in 39.82s`, including nonzero-base masked load/store coverage.

Additional checks performed:

- `git diff --check`
- A5 offline compilation for both cache signatures completed successfully.

### Offset-expression support boundary

This section describes the interleave/deinterleave *specialized rewrite* in `InterleaveOptimization.cpp`, not whether a Triton pointer expression is legal in general. Let `B` denote the sum of non-constant `index` leaves exposed by an `arith.addi` tree, and let `C` denote the sum of its statically known integer leaves. A `failure()` from this specialization rejects only the rewrite: an unmasked load can take its generic lowering path, while a store pair remains on its original path. The old assertion below was different because it aborted compilation.

#### Before this PR

| Expression form | Specialized behavior |
| --- | --- |
| Static `0` or `1` | Accepted as the canonical even or odd offset. |
| Dynamic `B` with no immediately visible constant `arith.addi` | Treated as an even-base view and reused unchanged. |
| Canonical outer `B + 1` or `1 + B` | Intended odd-lane case, but the old backtrace required every constant `arith.addi` it encountered to also be `+1`. |
| Static `C` where `C != 0, 1`, or direct dynamic `B + C` where `C != 1` | Rejected by the shallow `checkIsCaseOffsetValid` gate; the specialization was not applied. |
| `((B + 64) + 1)` | The shallow gate accepted the outer `+1`, then the recursive backtrace reached `+64` and asserted. This is the reported compiler abort. |
| Repeated or embedded `+1` terms such as `(B + 1) + 1` | Not normalized as a total static offset; outside the old single-lane-marker model. |

#### With this PR

The new analysis recursively flattens only `arith.addi` and normalizes every analyzable expression as:

`B + C -> B + (C - (C mod 2)) + {EVEN_MODE, ODD_MODE}`

| Expression form | Normalized result / behavior |
| --- | --- |
| Static `C`, or dynamic `B + C`, where the accumulated `C` fits in `int64_t` | Supported by the specialization. The base constant is the pair-aligned `C - (C mod 2)` and the lane is selected from parity. |
| `B + 64` / `B + 65` | Normalize to the same base `B + 64`, with EVEN / ODD respectively. This is the FlagGems nonzero-base case. |
| Re-associated forms such as `((B + 64) + 1)`, `((B + 1) + 64)`, or `1 + (B + 64)` | The recursive collector accumulates the same static total `65`; each becomes base `B + 64` plus ODD. |
| Folded constant form `B + 65` | Also becomes base `B + 64` plus ODD. It may require materializing `B + 64` because no outer `+1` SSA edge is available to reuse. |
| A directly reusable even offset, or an odd offset whose outermost operation is `B + 1` / `1 + B` | Reuses the existing SSA base. Other valid forms are rebuilt canonically by `materializeDeinterleaveBaseOffset`. |
| A paired store `(B + C_even, B + C_odd)` | Optimized only when the normalized bases have the same static base constant and the same dynamic base-term multiset, while the modes differ. Addition order among exposed base terms may differ. |

#### Still outside the proven specialized-expression domain

- A non-`index` dynamic leaf, or static-constant accumulation that overflows `int64_t`, returns `failure()`.
- The analysis deliberately does not recurse through non-`arith.addi` producers such as multiply, addptr, affine/apply-style, or arbitrary base-address subgraphs. A constant hidden inside such a producer is kept inside an opaque base term; it is not inferred to be a lane selector.
- Pairing is based on exact SSA-value identity of the exposed base-term multiset, not algebraic equivalence. For example, `X + 64` and `Y + 65` do not form a pair when `X` and `Y` are distinct Values, even if another analysis could prove them mathematically equal.
- For stores, equal modes or different normalized bases decline the interleave rewrite. Existing shape, stride, mask, and candidate-discovery prerequisites remain unchanged.

Accordingly, the new boundary is intentionally broader for additive offset expressions and safer for everything else: unsupported or unprovable cases decline the specialization rather than asserting.

### Offset-analysis call flow and removed store branches

The new parsing and normalization helpers (`addStaticOffset`, `collectDeinterleaveOffsetTerms`, `normalizeDeinterleaveOffset`, `hasEquivalentDeinterleaveBase`, and `materializeDeinterleaveBaseOffset`) are file-local `static` functions. `expandInterleaveMemRefType` is an existing namespace helper; it receives the final normalized cast offset. The parsing helpers have no public API callers; their entry points are the four existing load/store specializations.

```mermaid
flowchart TD
  ULoad["Unmasked tt.load"] --> UConverter["LoadConverter::matchAndRewrite"] --> UOpt["DeinterleaveStatusOptimization"]
  MLoad["Masked tt.load (contiguous-mask checks pass)"] --> MConverter["LoadConverter::matchAndRewrite"] --> MOpt["DeinterleaveStatusWithMaskOptimization"]
  UStore["Unmasked paired stores"] --> UPass["TritonToLinalgPass groups two candidates"] --> UStoreOpt["InterleaveStatusOptimization"]
  MStore["Masked paired stores"] --> MPass["TritonToLinalgPass groups masked candidates"] --> MStoreOpt["InterleaveStatusWithMaskOptimization"]
```

Each load specialization has one original `reinterpret_cast` offset. Each store specialization has two offsets, one for each candidate lane. The shared internal flow is:

```mermaid
flowchart TD
  Origin["originCastOffset(s)"] --> Normalize["normalizeDeinterleaveOffset"]
  Normalize --> Collect["collectDeinterleaveOffsetTerms"]
  Collect --> Add["addStaticOffset for each exposed constant"]
  Collect --> Validate["validate index base terms"]
  Validate --> Info["baseConstant + indexMode + optional reusable SSA base"]
  Info --> Materialize["materializeDeinterleaveBaseOffset"]
  Materialize --> Expand["expandInterleaveMemRefType(type, castOffset)"]
  Expand --> Recast["create expanded reinterpret_cast"]
```

For loads, the resulting `indexMode` selects `extract_slice` lane 0 or 1. For stores, both offsets are first normalized; `hasEquivalentDeinterleaveBase` then requires the same normalized base and opposite modes before the two sources are inserted into lanes 0 and 1 of one expanded destination. `hasEquivalentDeinterleaveBase` is therefore store-only; it is not needed for a single load.

A failed unmasked-load specialization falls through to the generic load copy path. A failed paired-store specialization performs no rewrite, so the original materialization path remains. This is the intended non-assert fallback boundary.

#### Why the former `InterleaveStatusOptimization` offset branches were removed

The deleted block formerly split offset handling by representation:

- `Attribute` / `Attribute`: it stripped the modes and unconditionally chose static `newCastOffset = 0`.
- `Value` / `Value`: it stripped the modes, required the two residual offsets to be the same single SSA `Value`, then chose whichever residual belonged to the EVEN mode.
- `Attribute` / `Value`: it rejected the pair outright.

Those branches encoded the old assumption that the only static offset information was the lane marker `0` or `1`, and that a dynamic base could be represented by one raw residual `Value`. That is why they cannot represent the pair `B + 64` and `B + 65`: their shared base is `B + 64`, not `0`, and the `+64` must not be treated as a lane marker.

The replacement normalizes both operands before comparing them. It tests `baseConstant`, compares the exposed dynamic base terms as a multiset, requires opposite modes, and then materializes the common normalized base from either operand's equivalent `DeinterleaveOffsetInfo`. This removes the artificial static/dynamic split, preserves a nonzero base in both the new `reinterpret_cast` operand and its MemRef layout, and permits re-associated additive trees without accepting arbitrary algebraic equivalence.

### DIFF
--- a/third_party/ascend/lib/Utils/InterleaveOptimization.cpp
+++ b/third_party/ascend/lib/Utils/InterleaveOptimization.cpp
@@ -48,9 +48,9 @@ namespace mlir {
 namespace triton {
 // For origin MemRefType of ReinterpretCastOp under interleave state, here wanna
 // adjust its shape info by expanding last dimension double.
-MemRefType
-expandInterleaveMemRefType(MemRefType originType,
-                           std::optional<OpFoldResult> castOffset = std::nullopt) {
+MemRefType expandInterleaveMemRefType(
+    MemRefType originType,
+    std::optional<OpFoldResult> castOffset = std::nullopt) {
   // Double the last dimension shape
   SmallVector<int64_t> shape(originType.getShape());
   shape.back() = shape.back() * 2;
@@ -469,7 +469,7 @@ InterleaveStatusOptimization(SmallVector<Operation *> materializeVec) {
   // Create new op
   // 1. Get new destination memref type
   auto dstType = expandInterleaveMemRefType(firstReinterpretCastOp.getType(),
-                                             newCastOffset);
+                                            newCastOffset);
 
   // 2. New tensor::EmptyOp
   auto emptyTensor = builder.create<tensor::EmptyOp>(loc, dstType.getShape(),
@@ -619,7 +619,7 @@ InterleaveStatusWithMaskOptimization(SmallVector<Operation *> materializeVec) {
   // Create new op
   // 1. Get new destination memref type
   auto dstType = expandInterleaveMemRefType(firstReinterpretCastOp.getType(),
-                                             newCastOffset);
+                                            newCastOffset);
 
   // 2. New tensor::EmptyOp
   auto emptyTensor = builder.create<tensor::EmptyOp>(loc, dstType.getShape(),
--- a/third_party/ascend/lib/Utils/InterleaveOptimization.cpp
+++ b/third_party/ascend/lib/Utils/InterleaveOptimization.cpp
@@ -48,7 +48,7 @@ namespace mlir {
 namespace triton {
 // For origin MemRefType of ReinterpretCastOp under interleave state, here wanna
 // adjust its shape info by expanding last dimension double.
-MemRefType expandInterleaveMemRefType(
+static MemRefType expandInterleaveMemRefType(
     MemRefType originType,
     std::optional<OpFoldResult> castOffset = std::nullopt) {
   // Double the last dimension shape
--- a/third_party/ascend/lib/Utils/InterleaveOptimization.cpp
+++ b/third_party/ascend/lib/Utils/InterleaveOptimization.cpp
@@ -40,13 +40,17 @@
 #include "llvm/Support/Casting.h"
 #include "llvm/Support/Debug.h"
 #include <cassert>
+#include <limits>
+#include <optional>
 #include <utility>
 
 namespace mlir {
 namespace triton {
 // For origin MemRefType of ReinterpretCastOp under interleave state, here wanna
 // adjust its shape info by expanding last dimension double.
-MemRefType expandInterleaveMemRefType(MemRefType originType) {
+MemRefType
+expandInterleaveMemRefType(MemRefType originType,
+                           std::optional<OpFoldResult> castOffset = std::nullopt) {
   // Double the last dimension shape
   SmallVector<int64_t> shape(originType.getShape());
   shape.back() = shape.back() * 2;
@@ -54,10 +58,20 @@ MemRefType expandInterleaveMemRefType(MemRefType originType) {
   // Adjuest layout attribute
   StridedLayoutAttr originLayout =
       llvm::dyn_cast<StridedLayoutAttr>(originType.getLayout());
-  // If offset is static, just reset it to 0
+  // If offset is static, just reset it to 0.
   auto offset = originLayout.getOffset() == ShapedType::kDynamic
                     ? originLayout.getOffset()
                     : 0;
+  // A reinterpret_cast result type must agree with its offset operand.  The
+  // original interleaved view can carry layout offset 0 while the normalized
+  // base is a non-zero static offset, so use the new cast offset whenever it
+  // is available.
+  if (castOffset) {
+    if (auto staticOffset = getConstantIntValue(*castOffset))
+      offset = staticOffset.value();
+    else
+      offset = ShapedType::kDynamic;
+  }
   // Set last dimension stride to 1
   SmallVector<int64_t> stride(originLayout.getStrides());
   stride.back() = 1;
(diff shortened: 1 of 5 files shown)


third_party/ascend/lib/Utils/InterleaveOptimization.cpp
```diff
@@ -40,24 +40,38 @@
...
+#include <limits>
+#include <optional>
 #include <utility>
 
 namespace mlir {
 namespace triton {
 // For origin MemRefType of ReinterpretCastOp under interleave state, here wanna
 // adjust its shape info by expanding last dimension double.
-MemRefType expandInterleaveMemRefType(MemRefType originType) {
+MemRefType expandInterleaveMemRefType(
```
elstainniles: 这个在cpp里改动了参数，需要同步修改.h

zhanwei33: 已处理。全仓检索确认 `expandInterleaveMemRefType` 仅在 `InterleaveOptimization.cpp` 内部调用，不需要作为公共接口暴露。因此采用内部 helper 方案：将其改为 `static`，并从 `.h` 移除旧的单参数声明；同时清理了已删除的 `checkIsCaseOffsetValid` 和 `recountReinterpretCastOffset` 残留声明，避免声明与实现签名不一致。对应提交：8a689467。

WuTYSFG (approved):

HinPeng (approved):

PR #1419, opened 2026-08-06
[Libdevice](fix) Set libdevice simd operation always inline

Levein7:
1. Correct the name of local variables , avoid misleading the compilation mode.
2. Add "always_inline" attribute for libdevice operation in SIMD compilation mode.


hongziqi (approved):

KanuaK (approved):
lgtm

PR #1420, opened 2026-08-06
[FP_TO_FP](fix) Fix scalar RTNE fp-to-fp lowering

zhanwei33:
## Summary

- Route scalar RTNE/default `tt.fp_to_fp` operations past tensor-only HFusion conversion.
- Canonicalize scalar and ranked-tensor RTNE conversions in TritonToLinalg by using `getElementTypeOrSelf` and preserving `resultType`.

## FlagGems reproducer

The source workload is FlagGems `pack_seq` with an FP8 output:

- Historical tracked entry: `tests/test_pack_seq.py::test_pack_seq_fp8_basic[6-8-4-lengths_list0]`. It compiles `_pack_seq_kernel` and originally exposed the TA `Casting.h:566` assertion.
- `test_pack_seq_fp8_basic` now explicitly passes `pad_value=0.0`. It checks valid sequence elements only and is a smoke-test workaround for the later CCEC nonzero-FP8-literal limitation; it does not change the public default.
- The public/default and finite-nonzero forms are exercised by `test_pack_seq_fp8_default_inf_padding` and `test_pack_seq_fp8_custom_padding` (`-100.0`, `0.0`, `100.0`). The kernel always contains the padding-initialization statement at compile time, even where a particular runtime shape has no visible padded output region.

Representative commands from the FlagGems checkout:

```bash
pytest -q 'tests/test_pack_seq.py::test_pack_seq_fp8_basic[6-8-4-lengths_list0]'
pytest -q tests/test_pack_seq.py::test_pack_seq_fp8_custom_padding
pytest -q tests/test_pack_seq.py::test_pack_seq_fp8_default_inf_padding
```

After this PR, a nonzero/default FP8-padding compile can progress past the original TA assertion but may still stop in the separate CCEC literal-format issue (`float8e4m3 0xPFF` for `-inf`, or `0xPEC` for `-100.0`). Zero padding is a diagnostic/smoke workaround only, not a replacement for the public `-inf` default.

## Representative DSL and why the conversion is scalar

`pack_seq` passes `pad_value` as `PAD_VALUE: tl.constexpr`; for non-`uint8` output it builds an FP32 padding block and stores it through an output pointer whose dtype follows `x`:

```python
PAD_VALUE: tl.constexpr

# For FP8 x, out_ptr has FP8 element type.
pad_vals = tl.full([BLOCK_T, BLOCK_D], PAD_VALUE, tl.float32)
tl.store(out_row_ptr, pad_vals, mask=...)

# The later x_vals load/store is a separate tensor path.
x_vals = tl.load(x_row_ptr, mask=...)
tl.store(out_row_ptr, x_vals, mask=...)
```

The equivalent standalone reducer is:

```python
@triton.jit
def nonzero_fp8_constant_store(out_ptr, VALUE: tl.constexpr,
                               BLOCK: tl.constexpr):
    offsets = tl.arange(0, BLOCK)
    values = tl.full((BLOCK,), VALUE, tl.float32)
    tl.store(out_ptr + offsets, values)

# signature: out_ptr = *fp8e4nv; VALUE is a nonzero constexpr
```

`tl.full` means “broadcast one scalar value”; it does not make the source constant a tensor. Because the destination pointer element type is FP8, the compiler first converts the single FP32 constant to FP8, then splats that FP8 scalar to the block. The issue is therefore the padding `PAD_VALUE`, not a scalarization of the FP8 input tensor `x`.

## Corresponding TTIR

A captured A5 minimal specialization (using finite `VALUE = -100.0`; `-inf` has the same scalar conversion topology) contains:

```mlir
%cst = arith.constant -1.000000e+02 : f32
%offsets = tt.make_range {end = 64 : i32, start = 0 : i32} : tensor<64xi32>
%ptrs = tt.splat %out_ptr : !tt.ptr<f8E4M3FN> -> tensor<64x!tt.ptr<f8E4M3FN>>
%addrs = tt.addptr %ptrs, %offsets : tensor<64x!tt.ptr<f8E4M3FN>>, tensor<64xi32>
%fp8 = tt.fp_to_fp %cst, rounding = rtne : f32 -> f8E4M3FN
%values = tt.splat %fp8 : f8E4M3FN -> tensor<64xf8E4M3FN>
tt.store %addrs, %values : tensor<64x!tt.ptr<f8E4M3FN>>
```

The key facts are:

- `%cst` and `%fp8` are scalar types, so `tt.fp_to_fp` is legally scalar.
- `tt.splat` occurs **after** conversion and creates the tensor value required by `tt.store`.
- A tensor output requirement does not make the preceding conversion a tensor conversion. A lowering that unconditionally uses `cast<TensorType>` or `cast<RankedTensorType>` on this op is therefore invalid.

## Root cause

The pack-seq FP8 path creates this scalar `f32 -> f8E4M3FN` conversion followed by `tt.splat`. Both lowering patterns assumed `TensorType` or `RankedTensorType`, so MLIR `cast` asserted before the scalar conversion could lower.

## TritonToHFusion change

- Move the rounding-mode guard before tensor-only type inspection. For RTNE or an omitted rounding mode, the HFusion rewrite returns `failure()` immediately and leaves the op for TritonToLinalg.
- Keep `cast<TensorType>` after the rounding guard. This deliberately retains HFusion's tensor precondition: a scalar non-RTNE conversion is still rejected by the existing assertion rather than silently becoming a pattern miss.
- Keep the existing HFusion path for non-RTNE tensor conversions. The `tt.splat` following the scalar conversion is unchanged; it remains responsible for broadcasting the already-converted FP8 scalar to the tensor stored by the kernel.

`failure()` here means that this greedy rewrite pattern did not apply; it does not fail the HFusion pass. `applyPatternsGreedily` continues and later passes can lower the retained operation.

## Risk analysis and boundary

- RTNE/default tensor behavior is unchanged: it was already rejected by this HFusion pattern and handled by TritonToLinalg. The reordered guard simply lets scalar RTNE reach the same decision before any tensor-only cast.
- Non-RTNE tensor behavior is unchanged: it still reaches HFusion after the existing tensor precondition.
- The change only affects dispatch and type handling. It does not alter the requested rounding semantics, the scalar-to-tensor `tt.splat`, or the FP8 store layout.

## Unsupported conversion boundaries deliberately retained

The pass ownership remains:

| `tt.fp_to_fp` form | Intended owner |
| --- | --- |
| RTNE/default | TritonToLinalg |
| non-RTNE (for example RTZ) | TritonToHFusion |

- A scalar non-RTNE conversion does not take the RTNE early exit. It reaches `cast<TensorType>` in HFusion and retains the pre-existing assertion. This PR intentionally does **not** claim scalar non-RTNE support.
- A same-bitwidth conversion between distinct floating-point types (for example `f8E4M3FN -> f8E5M2`) remains rejected. It does not become an identity merely because both types are 8-bit; the Linalg equal-width path cannot replace an `f8E5M2` result with an `f8E4M3FN` value.

The scalar RTNE fix is therefore narrow: it only makes the legal `f32 -> FP8` RTNE/default scalar form reach its intended Linalg lowering.

## Validation

- `git diff --check` and targeted `pre-commit` passed on both branches.
- `TritonToHFusion.cpp` compiled with the project CMake-exported `clang++` command for both `main-dev` and `main` source revisions.
- New MLIR regression:

  ```bash
  triton-opt --triton-to-hfusion --triton-to-linalg \
    third_party/ascend/unittest/Conversion/General/TritonToLinalg/fp_to_fp_scalar_rtne.mlir \
    | FileCheck third_party/ascend/unittest/Conversion/General/TritonToLinalg/fp_to_fp_scalar_rtne.mlir
  ```

  It verifies scalar `f32 -> f8E4M3FN, rounding = rtne` lowers to `arith.truncf` with `round_mode = rint`.
- Direct offline A5 compilation used `triton.compile(..., target=GPUTarget("npu", "Ascend910_9589", 0))`; no NPU kernel was launched. The nonzero `-inf` repro passed the HFusion/Linalg boundary and emitted `arith.truncf`, then stopped only at the independent CCEC `float8e4m3 0xPFF` literal issue. The zero-padding specialization emitted `.npubin`.
- The FlagGems device test is not a compiler verdict in this host: on physical devices 0 and 1 it stops in `torch_npu` `x.to(torch.float8_e4m3fn)` with ACLNN error 561103 before JIT compilation starts.

### DIFF
--- a/third_party/ascend/lib/TritonToHFusion/TritonToHFusion.cpp
+++ b/third_party/ascend/lib/TritonToHFusion/TritonToHFusion.cpp
@@ -97,10 +97,11 @@ struct TritonFpToFpToHFusionConversion : OpRewritePattern<triton::FpToFpOp> {
       return failure();
     }
 
-    // Only non-RTNE tensor conversions are lowered to HFusion.
-    auto srcType = dyn_cast<TensorType>(input.getType());
-    auto dstType = dyn_cast<TensorType>(resultType);
-    if (!srcType || !dstType || !srcType.getElementType().isIntOrFloat() ||
+    // HFusion owns only non-RTNE tensor conversions. Keep the tensor
+    // precondition checked here; a scalar must not silently fall through.
+    auto srcType = cast<TensorType>(input.getType());
+    auto dstType = cast<TensorType>(resultType);
+    if (!srcType.getElementType().isIntOrFloat() ||
         !dstType.getElementType().isIntOrFloat()) {
       return failure();
     }
--- a/third_party/ascend/lib/TritonToHFusion/TritonToHFusion.cpp
+++ b/third_party/ascend/lib/TritonToHFusion/TritonToHFusion.cpp
@@ -100,8 +100,7 @@ struct TritonFpToFpToHFusionConversion : OpRewritePattern<triton::FpToFpOp> {
     // Only non-RTNE tensor conversions are lowered to HFusion.
     auto srcType = dyn_cast<TensorType>(input.getType());
     auto dstType = dyn_cast<TensorType>(resultType);
-    if (!srcType || !dstType ||
-        !srcType.getElementType().isIntOrFloat() ||
+    if (!srcType || !dstType || !srcType.getElementType().isIntOrFloat() ||
         !dstType.getElementType().isIntOrFloat()) {
       return failure();
     }
--- a/third_party/ascend/lib/TritonToHFusion/TritonToHFusion.cpp
+++ b/third_party/ascend/lib/TritonToHFusion/TritonToHFusion.cpp
@@ -87,21 +87,22 @@ struct TritonFpToFpToHFusionConversion : OpRewritePattern<triton::FpToFpOp> {
     Value input = op.getSrc();
     auto resultType = op.getResult().getType();
 
-    // Only handle float-to-float conversions with non-RTNE rounding modes
-    // RTNE (default) rounding is handled by TritonToLinalg pass using
-    // arith.truncf/extf
-    auto srcType = cast<TensorType>(input.getType());
-    auto dstType = cast<TensorType>(resultType);
-    if (!srcType.getElementType().isIntOrFloat() ||
-        !dstType.getElementType().isIntOrFloat()) {
-      return failure();
-    }
-
-    // Check if this has a non-RTNE rounding mode
+    // RTNE (default) rounding is handled by TritonToLinalg using
+    // arith.truncf/extf. Check this before querying tensor-only properties:
+    // scalar fp_to_fp operations are valid Triton IR too.
     auto roundingMode = op.getRounding();
     if (!roundingMode.has_value() ||
         roundingMode.value() == triton::RoundingMode::RTNE) {
       // RTNE or no rounding mode specified: let TritonToLinalg handle it
+      return failure();
+    }
+
+    // Only non-RTNE tensor conversions are lowered to HFusion.
+    auto srcType = dyn_cast<TensorType>(input.getType());
+    auto dstType = dyn_cast<TensorType>(resultType);
+    if (!srcType || !dstType ||
+        !srcType.getElementType().isIntOrFloat() ||
+        !dstType.getElementType().isIntOrFloat()) {
       return failure();
     }
 
--- a/third_party/ascend/include/TritonToLinalg/TritonOpConverter.h
+++ b/third_party/ascend/include/TritonToLinalg/TritonOpConverter.h
@@ -74,9 +74,14 @@ public:
 Convert `tt.fp_to_fp` operation with RTNE (default) rounding mode to
 `arith.truncf` or `arith.extf` operation.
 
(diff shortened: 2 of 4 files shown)


third_party/ascend/lib/TritonToHFusion/TritonToHFusion.cpp
```diff
@@ -87,24 +87,24 @@ struct TritonFpToFpToHFusionConversion : OpRewritePattern<triton::FpToFpOp> {
...
-    // arith.truncf/extf
-    auto srcType = cast<TensorType>(input.getType());
-    auto dstType = cast<TensorType>(resultType);
-    if (!srcType.getElementType().isIntOrFloat() ||
-        !dstType.getElementType().isIntOrFloat()) {
-      return failure();
-    }
-
-    // Check if this has a non-RTNE rounding mode
+    // RTNE (default) rounding is handled by TritonToLinalg using
```
hongziqi, line 90: It is recommended that some related test cases be added, such as end-to-end or MLIR tests.

zhanwei33:
- [ ]

PR #1424, opened 2026-08-06
[TritonToStructured](fix) Preserve zero-stride pointer broadcasts

zhanwei33:
## Summary

Fix the `TritonToStructured` lowering of a `tt.store` whose pointer is the
canonical broadcast form created for a non-singleton zero-stride dimension.

The triggering FlagGems case is
`tests/test_tril.py::test_tril_inplace_expanded_view`. Expanding `(1, 4)` to
`(3, 4)` gives the kernel `STRIDE_M=0`, `STRIDE_N=1`. Its pointer path is:

```text
tensor<1x64x!tt.ptr<f32>> --tt.broadcast--> tensor<16x64x!tt.ptr<f32>>
```

Previously `StoreConverter` rebuilt the pointer and mask with the broadcasted
zero-stride axis removed, while the store value remained `16x64`. The rebuilt
`tt.store` therefore failed verification because its value and pointer types
no longer matched.

## Design

1. Add a shared structural predicate,
   `ConverterUtils::isHoistablePointerBroadcast`, for a ranked pointer
   broadcast with exactly one broadcast dimension and an
   `addptr* -> splat(pointer)` source chain.
2. Let `StoreConverter` defer this legal pointer-broadcast store instead of
   rebuilding it. The existing `TritonToLinalg` broadcast hoister can then
   lower the form with its full rank intact.
3. Make `BroadcastHoister` use the same structural predicate, so the early
   hand-off and later hoist use one definition of the pointer-broadcast form.

This deliberately does not broaden generic zero-stride lowering: only the
form already targeted by the dedicated pointer-broadcast hoister is deferred.

## Predicate scope and pass hand-off

The shared predicate is a hand-off guard, not a statement that
`StoreConverter` and the hoister must globally match the same TTIR set. Let:

```text
D = stores that StoreConverter defers
H = pointer broadcasts that the later hoister can successfully lower
```

The required invariant is `D ⊆ H`: the early store rewrite may decline only
when a later path can take ownership. The reverse inclusion is neither
required nor desirable. `StoreConverter` only examines a `StoreOp` whose
direct pointer operand is a broadcast, while the hoister may see pointer
broadcasts with other uses.

`processPtrBroadcastOperations` also uses `BroadcastHoister::canBroadcast()`
to decide which pointer broadcasts are illegal in its partial conversion and
therefore must be consumed by the hoister. A real narrowing of that predicate
could affect non-store TTIR by routing it to the later generic
`TTOpConverters::BroadcastConverter` path. Such a change needs non-store
pointer-broadcast coverage.

For verifier-valid `tt.broadcast` IR, this refactoring is not intended to
narrow the old accepted set. The Triton verifier already requires equal rank
and requires every dimension either to have source extent `1` or to be
unchanged. The old `findSrc()` logic and the new helper also accept the same
`addptr* -> splat(pointer)` root chain. Making those conditions explicit is
useful because the predicate is now called directly from `StoreConverter`,
rather than relying on the old hoister's caller and constructor state.

## Known preflight limitation

`BroadcastHoister::parseAddptr()` additionally requires every `tt.addptr`
offset to be a ranked tensor with a static shape. The shared predicate does
not currently inspect the offsets along the `addptr*` chain. Consequently,
its `true` result means that the broadcast is structurally eligible; it is
not yet a proof that the complete hoist rewrite cannot fail.

This gap existed in the old `canBroadcast()` path as well, but it is more
important once the result also controls whether `StoreConverter` defers a
store. A follow-up should make the predicate a complete preflight by checking
each addptr offset's ranked/static condition, then reuse that result in the
store defer, hoister legality, and hoister rewrite. Before doing so, the
fallback behavior for non-eligible dynamic-offset IR must be validated; it
must not simply move a failure to a later conversion stage.

## Supported DSL scope

This change is intentionally narrow. It newly unblocks a `tl.store` whose
pointer lowers to the canonical single-axis pointer-broadcast form:

```text
splat(pointer) -> addptr* -> broadcast
tensor<...x1x...x!tt.ptr<T>> -> tensor<...xKx...x!tt.ptr<T>>, K != 1
```

At the DSL level, the typical case is a zero-stride `torch.expand` view used
as the pointer argument together with a statically shaped Triton tile:

```python
# Host: view.stride() == (0, 1)
view = base.expand(M, N)

offs_m = tl.arange(0, BLOCK_M)[:, None]
offs_n = tl.arange(0, BLOCK_N)[None, :]
offsets = offs_m * STRIDE_M + offs_n * STRIDE_N  # STRIDE_M == 0
tl.store(ptr + offsets, values, mask=mask)
```

The matcher/lowering accepts the form only when:

- the broadcast result is a ranked pointer tensor;
- source and result have the same rank, with exactly one `1 -> non-1`
  broadcast dimension and all other dimensions unchanged;
- tracing the source through zero or more `tt.addptr` operations reaches
  `tt.splat(pointer)`.

The dedicated hoister additionally requires every `addptr` offset to be a
ranked tensor with a static tile shape. That condition is met by the
regression and is required by `parseAddptr()`, but is not yet preflighted by
the shared predicate; see the known limitation above. It is a shape
requirement, not a constant-value requirement: offset values and masked
logical extents may still be runtime-dependent.

The end-to-end regression covers the row-zero-stride case
`1x64 -> 16x64`, corresponding to an expanded `(1, 4) -> (3, 4)` view.
The same matcher structurally covers a single broadcast axis in a higher-rank
static tile, subject to the rest of the lowering pipeline.

Aliased zero-stride stores still require meaningful write semantics: if
multiple logical elements target the same address, they must not write
conflicting values. The regression writes zero from every aliased row to make
the result deterministic.

## Out of scope

This PR does not claim generic zero-stride support. In particular, it does
not add support for:

- multiple broadcast/zero-stride dimensions, such as `1x1 -> MxN`;
- rank-changing pointer broadcasts;
- non-pointer broadcasts (for values or masks);
- pointer chains that are not `addptr* -> splat(pointer)`;
- dynamically shaped `addptr` offset tiles;
- arbitrary overlapping `as_strided` views or unrelated strided layouts;
- new `tl.load`, atomic, or other memory-operation support beyond this
  `tl.store` conversion path.

## Regression cases

- New MLIR regression:
  `third_party/ascend/unittest/Conversion/General/TritonToStructured/zero_stride_store_broadcast.mlir`
  checks that the legal `1x64 -> 16x64` pointer broadcast and matching store
  survive the Structured conversion.
- New native NPU end-to-end regression:
  `third_party/ascend/unittest/pytest_ut/test_zero_stride_store_broadcast.py`.
  - Expands a `(1, 4)` fp32 tensor to `(3, 4)`, preserving `stride=(0, 1)`.
  - Executes a `16x64` full-rank pointer/value/mask store through the expanded
    view, then checks the backing tensor, expanded view, stride, and data pointer.
  - Every aliased row writes the same zero value, making the result deterministic
    while preserving the pointer-broadcast shape that triggered the bug.
  - It has no architecture override and runs on the NPU selected by PyTorch, so
    the same node is usable on both A3 and A5 devices.
- FlagGems integration:
  `tests/test_tril.py::test_tril_inplace_expanded_view`.
- Offline Ascend 91095 compile matrix: zero `STRIDE_M`, shapes `(3, 4)`,
  `(16, 64)`, `(17, 65)`, diagonals `-1/0/1`, `debug=False/True`, plus a
  non-zero-stride control. This is 19 direct-compilation cases in total.

### MLIR regression

```bash
triton-opt --triton-to-structured="enable-mask-fallback-conversion=false optimize-dynamic-offset=false" \
  third_party/ascend/unittest/Conversion/General/TritonToStructured/zero_stride_store_broadcast.mlir \
  | FileCheck third_party/ascend/unittest/Conversion/General/TritonToStructured/zero_stride_store_broadcast.mlir
```

### A3/A5 native runtime regression

Run this unchanged on an A3 or A5 host after installing the PR wheel:

```bash
source /usr/local/Ascend/driver/bin/setenv.bash
source /home/yixuan/triton/pkg/cann9.0.0/ascend-toolkit/set_env.sh
python -m pytest -q -s \
  third_party/ascend/unittest/pytest_ut/test_zero_stride_store_broadcast.py
```

### FlagGems runtime integration

```bash
source /usr/local/Ascend/driver/bin/setenv.bash
source /home/yixuan/triton/pkg/cann9.0.0/ascend-toolkit/set_env.sh
ASCEND_RT_VISIBLE_DEVICES=3 \
PYTHONPATH=/home/w00609825/triton_workspace/FlagGems_new/src \
python -m pytest -q tests/test_tril.py::test_tril_inplace_expanded_view --ref cpu
```

### Explicit A5 offline compilation

The local matrix harness imports FlagGems'
`_tril_inplace_zero_strided_tile_kernel` and calls only
`triton.compile(ASTSource(...), target=GPUTarget("npu", "Ascend910_9589", 0))`:

```bash
source /usr/local/Ascend/driver/bin/setenv.bash
source /home/yixuan/triton/pkg/cann9.0.0/ascend-toolkit/set_env.sh
PYTHONPATH=/home/w00609825/triton_workspace/FlagGems_new/src \
TRITON_ALWAYS_COMPILE=1 TRITON_KERNEL_DUMP=1 \
TRIL_A5_VALIDATION_DIR=<output-dir> \
python <local-validation-dir>/compile_flaggems_tril_a5.py
```

The harness does not create a launcher, load a binary, or invoke a kernel.

## Local results

- The new MLIR regression passed on both target-branch worktrees.
- All 19 explicit A5 compile cases emitted `ttir`, `ttadapter`, `mlirbc`,
  `bcmlir`, and `npubin`.
- The new native runtime node passed on the available Ascend910B4 (A3) device:
  `1 passed in 1.89s`.
- The same new specialization also produced a 215,512-byte `Ascend910_9589`
  `.npubin` through strict direct compilation without launch. This is A5
  code-generation evidence, not A5 native execution evidence.
- `test_tril_inplace_expanded_view --ref cpu`: `9 passed`.
- `test_01_vector_add.py::test_vector_addition`: `1 passed` as a TA smoke test.

## Captured passing TTIR → TTAdapter IR

The following matching excerpts are from one successful physical Ascend910B4
capture of
`tests/test_tril.py::test_tril_inplace_expanded_view[dtype0-0]`
for an expanded `(1, 4) -> (3, 4)` fp32 view
(`STRIDE_M=0, STRIDE_N=1`). They show the full-rank pointer-broadcast
store which this PR preserves and hands to the later lowering.

Only the all-upper-tile store is shown below; the diagonal-boundary store has
the same pointer chain and is lowered through the same hand-off.

### TTIR: canonical zero-stride pointer broadcast

```mlir
%2 = tt.splat %ptr
    : !tt.ptr<f32> -> tensor<1x64x!tt.ptr<f32>>
%3 = tt.addptr %2, %offs_n_8
    : tensor<1x64x!tt.ptr<f32>>, tensor<1x64xi32>
%4 = tt.broadcast %3
    : tensor<1x64x!tt.ptr<f32>> -> tensor<16x64x!tt.ptr<f32>>
tt.store %4, %cst, %mask_13
    : tensor<16x64x!tt.ptr<f32>>
```

This satisfies the shared predicate exactly:

```text
tt.broadcast:  tensor<1x64xptr> -> tensor<16x64xptr>
source chain:  tt.broadcast <- tt.addptr <- tt.splat(scalar pointer)
```

It is a same-rank pointer broadcast with one expanded dimension
(`1 -> 16`), all other dimensions unchanged, and the required
`addptr* -> splat(pointer)` root chain. Therefore `StoreConverter`
defers this `tt.store` instead of rebuilding a lower-rank pointer/mask.

### Actual pointer-broadcast hoist

`isHoistablePointerBroadcast` is only the eligibility guard. Once it returns
true, `processPtrBroadcastOperations` routes the pointer broadcast to
`BroadcastConverter`, which invokes `BroadcastHoister::parse()` and
`replaceBroadcastOp()`. The actual rewrite is:

```text
broadcast(addptr(splat(ptr), offs_1x64))
        ↓
addptr(splat(ptr)_16x64, broadcast(offs_1x64)_16x64)
```

Thus the original broadcast after pointer arithmetic is not merely ignored:
the hoister recreates the scalar-pointer splat at full tile rank, broadcasts
the `1x64` offset to `16x64`, creates a full-rank `tt.addptr`, and
replaces the original pointer `tt.broadcast`. The subsequent generic
Triton-to-Linalg lowering then converts that full-rank pointer/store form to
the `scf.for` and `memref` operations shown below.

### Corresponding TTAdapter IR: explicit physical writes

The later `TritonToLinalg` lowering consumes that preserved form. The
irrelevant dynamic boundary-size calculations are omitted here; the shown
operations are the direct lowering of the store above.

```mlir
scf.for %arg9 = %c0 to %12 step %c1 {
  %reinterpret_cast = memref.reinterpret_cast %ptr
      to offset: [%offs_n_5], sizes: [1, 64], strides: [64, 1]
      : memref<?xf32>
        to memref<1x64xf32, strided<[64, 1], offset: ?>>

  %extracted_slice = tensor.extract_slice %3[0, 0] [%23, %25] [1, 1]
      : tensor<1x64xf32> to tensor<?x?xf32>
  %subview = memref.subview %reinterpret_cast[0, 0] [%23, %25] [1, 1]
      : memref<1x64xf32, strided<[64, 1], offset: ?>>
        to memref<?x?xf32, strided<[64, 1], offset: ?>>
  bufferization.materialize_in_destination %extracted_slice in writable %subview
      : (tensor<?x?xf32>,
         memref<?x?xf32, strided<[64, 1], offset: ?>>) -> ()
}
```

Thus TTAdapter does not retain a literal `16x64` pointer broadcast. It
represents the logical M-dimension broadcast with a loop and a repeated
`1x64` memref view. Since `STRIDE_M=0`, every logical-row iteration
addresses the same physical row; value/mask semantics remain preserved rather
than being silently reduced to rank one.

### DIFF
--- a/third_party/ascend/include/Utils/Utils.h
+++ b/third_party/ascend/include/Utils/Utils.h
@@ -91,11 +91,6 @@ SmallVector<int64_t> getBroadcastDims(RankedTensorType src,
 SmallVector<int64_t> getUnbroadcastDims(RankedTensorType src,
                                         RankedTensorType dst);
 
-// Return true when a pointer broadcast can be lowered by the dedicated
-// pointer-broadcast hoister. Keep this predicate shared with earlier memory
-// rewrites so they do not destroy a legal form expected by that lowering.
-bool isHoistablePointerBroadcast(triton::BroadcastOp op);
-
 } // namespace ConverterUtils
 
 class ConversionPatternRewriter;
--- a/third_party/ascend/include/Utils/Utils.h
+++ b/third_party/ascend/include/Utils/Utils.h
@@ -91,6 +91,11 @@ SmallVector<int64_t> getBroadcastDims(RankedTensorType src,
 SmallVector<int64_t> getUnbroadcastDims(RankedTensorType src,
                                         RankedTensorType dst);
 
+// Return true when a pointer broadcast can be lowered by the dedicated
+// pointer-broadcast hoister. Keep this predicate shared with earlier memory
+// rewrites so they do not destroy a legal form expected by that lowering.
+bool isHoistablePointerBroadcast(triton::BroadcastOp op);
+
 } // namespace ConverterUtils
 
 class ConversionPatternRewriter;
--- a/third_party/ascend/lib/TritonToStructured/MemOpConverter.cpp
+++ b/third_party/ascend/lib/TritonToStructured/MemOpConverter.cpp
@@ -295,6 +295,17 @@ LogicalResult StoreConverter::matchAndRewrite(triton::StoreOp op,
       return success();
   }
 
+  // AddPtrSplatConverter can preserve a non-singleton zero-stride dimension
+  // as a pointer broadcast. TritonToLinalg has a dedicated lowering for this
+  // legal form; rebuilding it here drops the zero-stride dimension from the
+  // pointer and mask while the value remains full-rank.
+  if (auto ptrBroadcast = oldPtr.getDefiningOp<triton::BroadcastOp>();
+      ptrBroadcast &&
+      ConverterUtils::isHoistablePointerBroadcast(ptrBroadcast)) {
+    return rewriter.notifyMatchFailure(
+        op, "defer hoistable pointer broadcast store to TritonToLinalg");
+  }
+
   MemOpTransformer tf(MemOpTransformer::MemType::store, optimizeDynamicOffset);
 
   auto newPtr = tf.createNewPtr(oldPtr, loc, rewriter);
--- a/third_party/ascend/lib/TritonToStructured/MemOpConverter.cpp
+++ b/third_party/ascend/lib/TritonToStructured/MemOpConverter.cpp
@@ -226,9 +226,8 @@ LogicalResult rewriteCatPointerStore(triton::StoreOp op, triton::CatOp catOp,
 
 // TTIR requires a store value to have the same shape as its pointer. A DSL
 // scalar is consequently represented either by tt.splat of a scalar or by a
-// folded dense splat constant. Only these uniform values are safe to defer for
-// a zero-stride pointer broadcast: all aliased lanes then write the same value.
-bool isScalarStoreValue(Value value) {
+// folded dense splat constant.
+bool isUniformScalarStoreValue(Value value) {
   if (auto splat = value.getDefiningOp<triton::SplatOp>())
     return !isa<ShapedType>(splat.getSrc().getType());
 
@@ -237,6 +236,46 @@ bool isScalarStoreValue(Value value) {
       return dense.isSplat();
   }
   return false;
+}
+
+// Return true only for the store form that must be left to the later pointer
+// broadcast lowering. A non-singleton zero-stride pointer dimension aliases
+// lanes, so deferring is sound only when the DSL supplied one scalar value.
+bool shouldDeferScalarPointerBroadcastStore(Value ptr, Value value) {
+  if (!isUniformScalarStoreValue(value))
+    return false;
+
+  auto broadcast = ptr.getDefiningOp<triton::BroadcastOp>();
+  if (!broadcast)
+    return false;
+
+  auto resultType = dyn_cast<RankedTensorType>(broadcast.getType());
+  auto sourceType = dyn_cast<RankedTensorType>(broadcast.getSrc().getType());
+  if (!resultType || !sourceType ||
+      !isa<triton::PointerType>(resultType.getElementType()) ||
+      resultType.getRank() != sourceType.getRank()) {
+    return false;
+  }
+
+  int broadcastedDims = 0;
+  for (size_t i = 0; i < resultType.getShape().size(); ++i) {
+    const int64_t srcDim = sourceType.getShape()[i];
+    const int64_t resultDim = resultType.getShape()[i];
+    if (srcDim == 1 && resultDim != 1) {
+      ++broadcastedDims;
+      continue;
+    }
+    if (srcDim != resultDim)
+      return false;
+  }
+  if (broadcastedDims != 1)
+    return false;
+
+  Value source = broadcast.getSrc();
+  while (auto addPtr = source.getDefiningOp<triton::AddPtrOp>())
+    source = addPtr.getPtr();
+  auto splat = source.getDefiningOp<triton::SplatOp>();
+  return splat && isa<triton::PointerType>(splat.getSrc().getType());
 }
 
 } // namespace
@@ -311,15 +350,11 @@ LogicalResult StoreConverter::matchAndRewrite(triton::StoreOp op,
   }
 
   // AddPtrSplatConverter can preserve a non-singleton zero-stride dimension
-  // as a pointer broadcast. TritonToLinalg has a dedicated lowering for this
-  // legal form; rebuilding it here drops the zero-stride dimension from the
-  // pointer and mask while the scalar store value remains full-rank.
-  if (auto ptrBroadcast = oldPtr.getDefiningOp<triton::BroadcastOp>();
-      ptrBroadcast &&
-      isScalarStoreValue(oldValue) &&
-      ConverterUtils::isHoistablePointerBroadcast(ptrBroadcast)) {
+  // as a pointer broadcast. Rebuilding the scalar store here drops that
+  // dimension from the pointer and mask while the store value stays full-rank.
+  if (shouldDeferScalarPointerBroadcastStore(oldPtr, oldValue)) {
     return rewriter.notifyMatchFailure(
-        op, "defer hoistable pointer broadcast store to TritonToLinalg");
+        op, "defer scalar pointer broadcast store to TritonToLinalg");
   }
 
   MemOpTransformer tf(MemOpTransformer::MemType::store, optimizeDynamicOffset);
--- a/third_party/ascend/lib/TritonToStructured/MemOpConverter.cpp
+++ b/third_party/ascend/lib/TritonToStructured/MemOpConverter.cpp
@@ -224,6 +224,21 @@ LogicalResult rewriteCatPointerStore(triton::StoreOp op, triton::CatOp catOp,
   return success();
 }
 
+// TTIR requires a store value to have the same shape as its pointer. A DSL
+// scalar is consequently represented either by tt.splat of a scalar or by a
+// folded dense splat constant. Only these uniform values are safe to defer for
+// a zero-stride pointer broadcast: all aliased lanes then write the same value.
+bool isScalarStoreValue(Value value) {
+  if (auto splat = value.getDefiningOp<triton::SplatOp>())
+    return !isa<ShapedType>(splat.getSrc().getType());
+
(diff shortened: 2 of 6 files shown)


third_party/ascend/include/Utils/Utils.h
```diff
@@ -91,6 +91,11 @@ SmallVector<int64_t> getBroadcastDims(RankedTensorType src,
 SmallVector<int64_t> getUnbroadcastDims(RankedTensorType src,
                                         RankedTensorType dst);
 
+// Return true when a pointer broadcast can be lowered by the dedicated
+// pointer-broadcast hoister. Keep this predicate shared with earlier memory
+// rewrites so they do not destroy a legal form expected by that lowering.
+bool isHoistablePointerBroadcast(triton::BroadcastOp op);
```
zhanwei33: Non-blocking rationale: this is intentionally a shared eligibility predicate, not a lowering API. `StoreConverter` must defer only the exact canonical form that the later hoister can reconstruct; centralizing the contract prevents an early rewrite from leaving a legal store to an unsupported lowering path.

third_party/ascend/lib/Utils/Utils.cpp
```diff
@@ -793,6 +793,38 @@ SmallVector<int64_t> getUnbroadcastDims(RankedTensorType src,
...
+    }
+    if (srcDim != resultDim) {
+      return false;
+    }
+  }
+  if (broadcastedDims != 1) {
+    return false;
+  }
+
+  Value source = op.getSrc();
```
zhanwei33: Non-blocking rationale: this def-use walk peels only the pointer operand of each `tt.addptr`, not its offset. It proves that the outer broadcast ultimately originates from one scalar pointer expanded by `tt.splat`; offsets may still vary per element and are rebuilt separately. Arbitrary pointer-tensor producers are intentionally excluded so the deferred store is always reconstructible by the hoister.

zhanwei33 (commented):
Non-blocking rationale notes for the shared pointer-broadcast predicate.

third_party/ascend/lib/TritonToStructured/MemOpConverter.cpp
```diff
@@ -295,6 +295,17 @@ LogicalResult StoreConverter::matchAndRewrite(triton::StoreOp op,
...
+  // AddPtrSplatConverter can preserve a non-singleton zero-stride dimension
+  // as a pointer broadcast. TritonToLinalg has a dedicated lowering for this
+  // legal form; rebuilding it here drops the zero-stride dimension from the
+  // pointer and mask while the value remains full-rank.
+  if (auto ptrBroadcast = oldPtr.getDefiningOp<triton::BroadcastOp>();
+      ptrBroadcast &&
+      ConverterUtils::isHoistablePointerBroadcast(ptrBroadcast)) {
+    return rewriter.notifyMatchFailure(
+        op, "defer hoistable pointer broadcast store to TritonToLinalg");
+  }
```
elstainniles, line 358: Ensure that the value of store must be a scalar

KanuaK (approved):

hongziqi (approved):

PR #1430, opened 2026-08-06
[TensorDescriptor](feat) support descriptor lowing to ptr

hongziqi:
## Summary
Enable Ascending kernels to accept Python `triton.tools.tensor_descriptor.TensorDescriptor` as arguments.

1. **`make_ttir`**: add `rewrite_tensor_pointer` + `rewrite_tensor_descriptor_to_pointer` so descriptor load/store become ordinary masked `tt.load` / `tt.store`.
2. **`driver.py`**: expand `tensordesc<...>` launch signature and unwrap host `TensorDescriptor` objects.
3. **`getScalarValue`**: peel `arith.select(padding, splat_nan, splat_zero)` produced by the rewrite, so `LoadConverter` can `linalg.fill` UB lanes.

## Why

Previously Ascend only exercised `tl.make_tensor_descriptor` **inside** the kernel. Passing an external `TensorDescriptor` failed with:

- `KeyError: 'tensordesc<...>'` in the launcher (signature not expanded), and/or
- `other value used in masked load produced by unsupported instruction` in TritonToLinalg (`other` was a tensor `select`).

### Pass roles

| Pass | Role |
|------|------|
| `triton-rewrite-tensor-descriptor-to-pointer` | `make_tensor_descriptor` / `descriptor_load|store` → ptr + shape/strides + padding, then masked load/store |

### Driver (vs NV)

Aligned with NV’s structure (`_expand_signature`, `_flatten_signature`, `make_tensordesc_arg`, `wrap_handle_tensordesc`), with Ascend-only deletions:

- No `tensordesc_meta` / `nvTmaDesc` / `CUtensorMap` (TMA-only on NV).
- Always take the pointer-decomposition branch.
- Filter packed kernel args with `ty != "constexpr"` (same idea as NV; no constant-index remap).

### `getScalarValue` (Utils.cpp)

Rewrite emits:

```mlir
%other = arith.select %padding, dense<nan>, dense<0> : tensor<Bm x Bn x T>
%tile  = tt.load %ptrs, %mask, %other
```

- Device `make_tensor_descriptor` with compile-time padding: canonicalize may fold the select → splat constant → already handled.
- **Host** `TensorDescriptor`: `%padding` is a function-arg `i1` → select cannot fold.

Peel path:

```text
select(i1, splat_nan, splat_zero)
  -> select(i1, scalar_nan, scalar_zero)
  -> scf.if { linalg.fill ins(%scalar_select) outs(%alloc) }
```

### Host copy (pytest)

```python
in_desc = TensorDescriptor(x, list(x.shape), list(x.stride()), [BLOCK_M, BLOCK_N])
out_desc = TensorDescriptor(y, list(y.shape), list(y.stride()), [BLOCK_M, BLOCK_N])
kernel[grid](in_desc, out_desc, M, N, BLOCK_M, BLOCK_N)
```

### After rewrite (TTIR sketch)

```mlir
// tensordesc arg -> ptr + shapes + strides + padding (+ host shape/stride dups)
%other = arith.select %padding, dense<nan>, dense<0> : tensor<16x32xf32>
%tile  = tt.load %ptrs, %mask, %other : tensor<16x32x!tt.ptr<f32>>
tt.store %out_ptrs, %tile, %mask
```

### After `triton-to-linalg` (host padding)

```mlir
%nan  = arith.constant 0x7FC00000 : f32
%zero = arith.constant 0.000000e+00 : f32
%pad  = arith.select %padding, %nan, %zero : f32
%alloc = memref.alloc() : memref<16x32xf32>
scf.if %need_pad {
  linalg.fill ins(%pad : f32) outs(%alloc : memref<16x32xf32>)
} {hivm.unlikely_condition}
memref.copy ...   // active region
bufferization.to_tensor %alloc
```

## CheckList
	- [ ] This PR does not need a test because …
	- [x] I have added tests.
		- `unittest/pytest_ut` for end-to-end NPU tests
		- `unittest/Conversion/General/TritonToLinalg` lit FileCheck
	- [x] The `lit` tests I have added follow FileCheck best practices (minimal TTIR kernels + documented lowering Q&A).

### DIFF
--- a/third_party/ascend/backend/driver.py
+++ b/third_party/ascend/backend/driver.py
@@ -183,7 +183,9 @@ class NPULauncher(object):
         spec = importlib.util.spec_from_file_location("__triton_launcher", self.so_launcher_path)
         mod = importlib.util.module_from_spec(spec)
         spec.loader.exec_module(mod)
-        self.launch = getattr(mod, "launch")
+        cst_key = lambda i: self.src.fn.arg_names.index(i) if isinstance(i, str) else i
+        signature = {cst_key(key): value for key, value in self.src.signature.items()}
+        self.launch = wrap_handle_tensordesc(getattr(mod, "launch"), signature)
 
     def _make_launcher_stub_path(self):
         header_src = generate_npu_header_src()
@@ -424,6 +426,8 @@ def generate_npu_header_src():
 def ty_to_cpp(ty):
     if ty[0] == '*':
         return "void*"
+    if ty.startswith("tensordesc"):
+        return "void*"
     return {
         "i1": "int32_t",
         "i8": "int8_t",
@@ -443,6 +447,34 @@ def ty_to_cpp(ty):
     }[ty]
 
 
+_BASE_ARGS_FORMAT = "iiiKKOOOO"
+_BASE_ARGS_FORMAT_LEN = len(_BASE_ARGS_FORMAT)
+
+
+def make_tensordesc_arg(arg):
+    return [arg.base, *arg.shape, *arg.strides, arg.padding == "nan", *arg.shape, *arg.strides]
+
+
+def wrap_handle_tensordesc(launcher, signature):
+    has_tensor_desc_arg = any(isinstance(sig, str) and sig.startswith("tensordesc") for sig in signature.values())
+    if not has_tensor_desc_arg:
+        return launcher
+
+    tensordesc_indices = set(
+        [i for i, sig in enumerate(signature.values()) if isinstance(sig, str) and sig.startswith("tensordesc")])
+
+    def inner(*args):
+        final_args = list(args[:_BASE_ARGS_FORMAT_LEN])
+        for i, arg in enumerate(args[_BASE_ARGS_FORMAT_LEN:]):
+            if i in tensordesc_indices:
+                final_args.extend(make_tensordesc_arg(arg))
+            else:
+                final_args.append(arg)
+        return launcher(*final_args)
+
+    return inner
+
+
 # the template is from triton-adapter HEAD. Wrapping the generated kernel binary into a python module
 def make_launcher(constants, signature, metadata):
     import os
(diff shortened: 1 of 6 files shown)


third_party/ascend/lib/Utils/Utils.cpp
```diff
@@ -210,33 +223,53 @@ Value getScalarValue(Value operand, Location loc,
...
     } else if (auto op = operand.getDefiningOp<arith::ConstantOp>()) {
       if (auto attr = dyn_cast<DenseElementsAttr>(op.getValue())) {
-        if (!attr.isSplat()) {
-          InFlightDiagnostic diag = emitError(loc)
-                                    << "other value used in masked load "
-                                       "produced by unsupported instruction";
+        if (!attr.isSplat())
           return nullptr;
-        }
         auto elemValue = attr.getSplatValue<Attribute>();
```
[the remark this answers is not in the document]
hongziqi: already fix

third_party/ascend/backend/driver.py
```diff
@@ -443,6 +447,34 @@ def ty_to_cpp(ty):
...
+_BASE_ARGS_FORMAT = "iiiKKOOOO"
+_BASE_ARGS_FORMAT_LEN = len(_BASE_ARGS_FORMAT)
+
+
+def make_tensordesc_arg(arg):
+    return [arg.base, *arg.shape, *arg.strides, arg.padding == "nan", *arg.shape, *arg.strides]
+
+
+def wrap_handle_tensordesc(launcher, signature):
+    has_tensor_desc_arg = any(isinstance(sig, str) and sig.startswith("tensordesc") for sig in signature.values())
```
zhanwei33, line 459: TensorDescriptor can appear inside a tuple or namedtuple: Triton’s native specializer recursively produces a signature such as ('tensordesc<fp32[16, 16]>', 'i32'), and the frontend recursively accepts that aggregate signature.
However, both has_tensor_desc_arg and _expand_signature only inspect top-level string entries. For an argument such as (desc, value), the wrapper is not installed and the descriptor leaf is never expanded into its pointer, shape, stride, and padding ABI fields. The generated launcher therefore packs the original Python object instead of the expanded descriptor ABI, while the descriptor-rewrite pass has already changed the kernel function signature to require the expanded fields.
This is not merely missing coverage: it produces a launcher/kernel ABI mismatch and can pass a host PyObject* where the device pointer is expected.
Please recursively traverse aggregate signatures and their runtime values, expanding TensorDescriptor leaves consistently with _flatten_signature, or explicitly reject nested TensorDescriptor arguments before compilation. Please also add an end-to-end regression for both tuple and namedtuple arguments containing a TensorDescriptor.

third_party/ascend/lib/Utils/Utils.cpp
```diff
@@ -210,33 +223,53 @@ Value getScalarValue(Value operand, Location loc,
       return reconstructScalarValue(operand);
     } else if (auto op = operand.getDefiningOp<arith::ConstantOp>()) {
       if (auto attr = dyn_cast<DenseElementsAttr>(op.getValue())) {
-        if (!attr.isSplat()) {
-          InFlightDiagnostic diag = emitError(loc)
-                                    << "other value used in masked load "
-                                       "produced by unsupported instruction";
+        if (!attr.isSplat())
```
zhanwei33, line 226: The new comment says that getScalarValue may return nullptr so that callers can fall back to a tensor-other path. There is currently no such fallback in LoadStoreConverter.
The only caller obtains scalarOther, checks it with an assert, and then unconditionally passes it to fillTensorWithOtherForMaskScenario. In release builds, NDEBUG removes the assertion, so an unsupported non-splat constant, select condition, or other unsupported producer can propagate an invalid Value into linalg.fill. This changes the previous behavior from a useful compiler diagnostic into a crash or invalid IR path.
Please either preserve the diagnostic before returning nullptr, or make the caller explicitly return a conversion failure when scalarization fails. The implementation should not silently return nullptr until a real tensor-other lowering path exists. A regression test should cover an unsupported non-splat other value and verify that compilation fails cleanly with a diagnostic.

third_party/ascend/lib/Utils/Utils.cpp
```diff
@@ -210,33 +223,53 @@ Value getScalarValue(Value operand, Location loc,
...
+          } else {
+            return nullptr;
+          }
+        } else {
+          return nullptr;
+        }
+      }
+      if (isa<ShapedType>(cond.getType()))
+        return nullptr;
+      return rewriter.create<arith::SelectOp>(loc, trueVal.getType(), cond,
```
zhanwei33, line 261: The new arith::SelectOp branch returns the newly created scalar select directly. This bypasses reconstructScalarValue, which is responsible for replaying SIToFPOp, UIToFPOp, and TruncFOp operations accumulated while walking the original tensor expression.
For example, a masked fp16 load can have an other expression shaped like:
%selected = arith.select %padding, %nan_f32, %zero_f32 : tensor<...xf32>
%other = arith.truncf %selected : tensor<...xf32> to tensor<...xf16>
getScalarValue records the truncf, reaches the select, scalarizes both branches to f32, and returns an f32 scalar without replaying the pending truncation. The subsequent fill then receives an f32 value for an fp16 allocation, leading to invalid IR or incorrect lowering.
Please construct the scalar select first and pass its result through reconstructScalarValue, e.g.:
auto scalarSelect = rewriter.create<arith::SelectOp>(
    loc, trueVal.getType(), cond, trueVal, falseVal);
return reconstructScalarValue(scalarSelect.getResult());
Please add a lit regression for select + truncf (and ideally select + uitofp) so this type-preservation contract is tested.

PR #1492, opened 2026-08-11
[TritonToLinalg](fix) reset ReinterpretCastOp result type offset to 0

liwei-link:
## Description
When baking the pointer offset into the base address during the "Calculate size of PointerCastOp precisely" optimization, the new ReinterpretCastOp sets offset to 0 in its static_offsets parameter, but the result MemRefType's StridedLayoutAttr still carries the original non-zero offset. This inconsistency between the op offset (0) and the type offset (non-zero) is fixed by reconstructing the MemRefType with offset 0 in the strided layout.

## Reproduce
```python
import torch
import torch_npu  # noqa: F401

import triton
import triton.language as tl


@triton.jit
def _inttoptr_static_offset_kernel(
    block_table_ptrs,
    out,
    BLOCK_SIZE: tl.constexpr,
    BLOCK: tl.constexpr,
):
    offsets = tl.arange(0, BLOCK)

    # Load a raw i64 address and cast to pointer — creates tt.int_to_ptr.
    block_table_ptr = tl.load(block_table_ptrs).to(tl.pointer_type(tl.int32))

    # Scalar addptr with a constant offset — creates a ReinterpretCastOp
    # whose result type has a non-zero static offset in the strided layout.
    block_table_ptr = block_table_ptr + 1

    block_numbers = tl.load(block_table_ptr + offsets)
    slot_ids = block_numbers.to(tl.int64)
    tl.store(out + offsets, slot_ids)


if __name__ == '__main__':
    """Static offset (constant 1) baked into base address."""
    block = 8

    block_table_cpu = torch.arange(16, dtype=torch.int32)
    block_table = block_table_cpu.npu()
    out = torch.empty((block,), dtype=torch.int64, device="npu")

    block_table_ptrs = torch.tensor(
        [block_table.data_ptr()], dtype=torch.int64
    ).npu()

    _inttoptr_static_offset_kernel[(1,)](
        block_table_ptrs,
        out,
        BLOCK_SIZE=4,
        BLOCK=block,
    )
    torch.npu.synchronize()

    # offset 1 => skip block_table[0]
    expected = block_table_cpu[1 : 1 + block].to(torch.int64)
    assert torch.equal(out.cpu(), expected.cpu())

```
Output:
```bash
test.py:24:46: error: expected result type with offset = 0 instead of 1
    block_numbers = tl.load(block_table_ptr + offsets)
                                             ^
```

## Fix
Before constructing the new ReinterpretCastOp , extract the original MemRefType , check if its layout is a StridedLayoutAttr , and if so, rebuild the layout with offset 0 (preserving the original strides). The new MemRefType with the corrected layout is then passed to the builder.

# New contributor declaration

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests

### DIFF
--- a/third_party/ascend/lib/TritonToLinalg/TritonToLinalgPass.cpp
+++ b/third_party/ascend/lib/TritonToLinalg/TritonToLinalgPass.cpp
@@ -1206,12 +1206,27 @@ void TritonToLinalgPass::runOnOperation() {
       markOp->setAttr(hivm::AddressSpaceAttr::getMnemonic(),
                       {hivm::AddressSpaceAttr::get(rewriter.getContext(),
                                                    hivm::AddressSpace::GM)});
+
+      // update result offset
+      auto origResultType =
+          cast<MemRefType>(reinterpretCastOp.getResult().getType());
+      MemRefType newResultType = origResultType;
+      if (auto stridedLayout =
+              dyn_cast<StridedLayoutAttr>(origResultType.getLayout())) {
+        int64_t offset = stridedLayout.getOffset();
+        if (!ShapedType::isDynamic(offset)) {
+          auto newLayout = StridedLayoutAttr::get(rewriter.getContext(), 0,
+                                                  stridedLayout.getStrides());
+          newResultType = MemRefType::get(
+              origResultType.getShape(), origResultType.getElementType(),
+              newLayout, origResultType.getMemorySpace());
+        }
+      }
+
       rewriter.replaceOpWithNewOp<memref::ReinterpretCastOp>(
-          reinterpretCastOp,
-          cast<MemRefType>(reinterpretCastOp.getResult().getType()), newCastOp,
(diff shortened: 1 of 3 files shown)


third_party/ascend/lib/TritonToLinalg/TritonToLinalgPass.cpp
```diff
@@ -1213,12 +1206,27 @@ void TritonToLinalgPass::runOnOperation() {
       markOp->setAttr(hivm::AddressSpaceAttr::getMnemonic(),
                       {hivm::AddressSpaceAttr::get(rewriter.getContext(),
                                                    hivm::AddressSpace::GM)});
+
+      // update result offset
```
KanuaK, line 1210: Please add e2e UT cases to verify this change.

hongziqi (approved):

KanuaK (approved):

PR #1494, opened 2026-08-11
[TritonToGraph](feat) Add merge-concat-load-buffer pass

Popeyesxs:
https://github.com/triton-lang/triton-ascend/issues/1086
A masked tt.load lowers to a full-tile UB alloc, a linalg.fill with the padding value, and a memref.copy of only the in-bounds subview. When two such loads are concatenated by tl.cat, triton-to-linalg emits an extract_slice + insert_slice pair over two separate allocs of the same static shape, so the kernel pays for twice the UB it needs plus two whole-tile fills.

This pass recognises that shape and folds the two allocs into one. The rewrite only fires when the source copies land inside the insert region and the dest copies stay clear of it, which makes the merge a pure renaming of the second buffer onto the first.

The fills are dropped only when the union of the copy destinations provably covers every later read of the concatenated tensor. When the padding is still observable the allocs are merged but one fill is kept, and when the two fills disagree on the padding value the rewrite is abandoned entirely.

# New contributor declaration

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests

### DIFF
--- a/third_party/ascend/backend/compiler.py
+++ b/third_party/ascend/backend/compiler.py
@@ -213,6 +213,13 @@ def ttir_to_linalg(mod, metadata, opt, *, named_ops=False):
         ascend.passes.ttir.add_triton_to_structure(pm, enable_mask_fallback_conversion, optimize_dynamic_offset)
         ascend.passes.ttir.add_triton_to_linalg(pm, False, named_ops, enable_nd2nz_on_vector, enable_select_analysis,
                                                 compile_on_910_95)
+        # Restricted to 910_95/950. The merged buffer is written by two disjoint
+        # memref.copy ops, and on 910_9362 the generated code only makes the
+        # copy next to the surviving to_tensor visible, so the other half of the
+        # tile reads back the padding value. The IR is unchanged through
+        # bishengir-opt, so the loss happens in code generation.
+        if compile_on_910_95:
+            ascend.passes.ttir.add_merge_concat_load_buffer(pm)
         if metadata["enable_dynamic_cv_pipeline"]:
             metadata["set_workspace_multibuffer"] = 0
             metadata["enable_mixed_cv"] = True
--- a/third_party/ascend/lib/TritonToGraph/MergeConcatLoadBuffer.cpp
+++ b/third_party/ascend/lib/TritonToGraph/MergeConcatLoadBuffer.cpp
@@ -1,0 +1,576 @@
+/*
+ * Copyright (c) Huawei Technologies Co., Ltd. 2025. All rights reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+#include "TritonToGraph/Passes.h"
+
+#include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/Bufferization/IR/Bufferization.h"
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/Linalg/IR/Linalg.h"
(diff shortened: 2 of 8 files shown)


third_party/ascend/backend/compiler.py
```diff
@@ -213,6 +213,7 @@ def ttir_to_linalg(mod, metadata, opt, *, named_ops=False):
         ascend.passes.ttir.add_triton_to_structure(pm, enable_mask_fallback_conversion, optimize_dynamic_offset)
         ascend.passes.ttir.add_triton_to_linalg(pm, False, named_ops, enable_nd2nz_on_vector, enable_select_analysis,
                                                 compile_on_910_95)
+        ascend.passes.ttir.add_merge_concat_load_buffer(pm)
```
hongziqi: It is recommended to add relevant end‑to‑end test cases to `third_party/ascend/unittest/pytest_ut`.

Popeyesxs: actually， you are right

third_party/ascend/lib/TritonToGraph/MergeConcatLoadBuffer.cpp
```diff
@@ -0,0 +1,545 @@
...
+                            << infoA->alloc << "\n  " << infoB->alloc << "\n");
+
+    // Keep the earlier alloc so RAUW cannot create uses-before-def.
+    bool aIsEarlier = infoA->alloc->isBeforeInBlock(infoB->alloc);
+    memref::AllocOp survivor = aIsEarlier ? infoA->alloc : infoB->alloc;
+    memref::AllocOp dropped = aIsEarlier ? infoB->alloc : infoA->alloc;
+    const FillInfo &survivorFill = aIsEarlier ? infoA->fill : infoB->fill;
+    const FillInfo &droppedFill = aIsEarlier ? infoB->fill : infoA->fill;
+
+    if (fillDead) {
```
zhanwei33, line 537: Update: the latest `fillPrecedesCopies` guard appears to address the execution-order condition below. Please retain it and add this as a regression, because the original geometric proof alone accepted the counterexample below.

The optimization is safe only when each `fill` is an initialization that executes before the copies it initializes. For example, for two `memref<8xf32>` temporaries, this order is safe:
```
fill(A, -1)                       # A = [-1, -1, -1, -1, -1, -1, -1, -1]
copy([10, 20, 30, 40], A[0:4])   # A = [10, 20, 30, 40, -1, -1, -1, -1]

fill(B, -1)                       # B = [-1, -1, -1, -1, -1, -1, -1, -1]
copy([50, 60, 70, 80], B[4:8])   # B = [-1, -1, -1, -1, 50, 60, 70, 80]

out = insert_slice(A[0:4], B[0:4])
# out = [10, 20, 30, 40, 50, 60, 70, 80]
```
Every element later read from `out` has been overwritten by a copy, so eliminating the fills after merging the two buffers preserves the result.

The pre-guard implementation also accepted the opposite order:
```
copy([10, 20, 30, 40], A[0:4])   # A = [10, 20, 30, 40, ?, ?, ?, ?]
fill(A, -1)                       # A = [-1, -1, -1, -1, -1, -1, -1, -1]

copy([50, 60, 70, 80], B[4:8])   # B = [?, ?, ?, ?, 50, 60, 70, 80]
fill(B, -1)                       # B = [-1, -1, -1, -1, -1, -1, -1, -1]

out = insert_slice(A[0:4], B[0:4])
# original out = [-1, -1, -1, -1, -1, -1, -1, -1]
```
Here each late fill overwrites the preceding copy. The old matcher checked only that every copy preceded `toTensor` and that the two write boxes covered the later read; it did not prove `fill -> copy` ordering. It therefore set `fillDead`, erased both fills, and RAUWd B onto A. The rewritten execution becomes:
```
copy([10, 20, 30, 40], A[0:4])
copy([50, 60, 70, 80], A[4:8])   # B has been merged into A
# fills erased
# rewritten out = [10, 20, 30, 40, 50, 60, 70, 80]
```
This changes a valid all-padding result into concatenated source values. Thus write-region coverage alone is not a proof that the fills are dead; the pass must also prove that each relevant fill precedes its copies. The same missing ordering constraint made the old `keepSingleFill` branch order-sensitive. The new guard should reject this counterexample because each fill follows its own copy.

zhanwei33 (changes requested):
The new end-to-end coverage is helpful. I found one correctness blocker in the fill-elision/merge ordering.

third_party/ascend/lib/TritonToGraph/MergeConcatLoadBuffer.cpp
Popeyesxs: Thanks — the guard is retained, and your counterexample is now a regression.

`fillPrecedesCopies` is applied at two different strengths, because the two
branches need different things:

- **Per buffer**, in `collectAllocInfo`: a fill must precede every copy into its
  own buffer. A fill that follows them is what the concatenated value observes,
  so the buffer is not accepted as a load buffer at all and the rewrite is
  abandoned. This is what rejects the second ordering in your example, and it is
  what protects the `fillDead` branch.
- **Across both buffers**, in the `keepSingleFill` branch: after the merge the
  surviving fill initialises the shared buffer, so it must precede *all* copies,
  not only the ones that used to target its own buffer. The per-buffer check is
  not sufficient here — a fill can precede its own copy and still follow the
  other alloc's copy, and merging would then let it clobber data that copy had
  already written. `survivor`/`dropped` are now selected before the fill
  decision so the check can be made against `allCopies`.

Regressions added in `merge_concat_load_buffer.mlir`:

- `no_merge_late_fill` — your counterexample. The copy precedes the fill it would
  initialise and the read region is fully covered, so the pre-guard matcher would
  have set `fillDead` and erased both fills.
- `no_merge_fill_after_other_copy` — each fill precedes its own copies, but the
  surviving fill follows the other buffer's copy, covering the order sensitivity
  you noted in `keepSingleFill`.

The comparison uses `isBeforeInBlock` and requires the same block: a fill or copy
in a different region is rejected rather than reasoned about. For a fill wrapped
in the `scf.if` that `fillTensorWithOtherForMaskScenario` emits, the `scf.if` is
used as the comparison point.v

zhanwei33: ok

WuTYSFG (approved):

HinPeng (approved):

PR #1512, opened 2026-08-12
[TritonToLinalg](fix) support chained byte offsets

elstainniles:
Support chained tensor byte offsets across pointer bitcasts and preserve expand_dims/broadcast address propagation. Adds an MLIR regression for the Ascend950 path.


KanuaK (approved):
lgtm

hongziqi (approved):

PR #1538, opened 2026-08-13
[ssbuffer](feat) scalar v2c dep opclassifier merge

1801ZDL:
Support scalar V→C dependencies in the Dynamic CV pipeline so that scalar values (e.g. loop bounds and extracted elements) produced by VECTOR blocks and consumed by CUBE blocks are transferred through the SSBuffer scalar channel.

  Previously these scalars were classified as CUBE (or missed entirely), which either left VECTOR math chains behind in CUBE scopes or required the whole kernel to fall back. The new classification makes every scalar that CUBE cannot produce itself an external input of the CUBE block, and the existing `analyzeExternalInputs` path turns it into a proper V→C dependency.

  ## Changes

  - **OpClassifier**: `tensor.extract` of a tensor with a VECTOR-only producer (`math.floor`/`math.ceil`, `linalg.reduce`, `arith.select` on tensors) is now classified as VECTOR through the shared `shouldSkipCubeUpstream` predicate used by both CUBE-upstream BFS paths, so the extract stays in the VECTOR block and is directly referenced by the CUBE block.
  - **SplitDataflow / analyzeExternalInputs**: the cross-block scalar is detected as a CUBE-block external input defined by a VECTOR op and recorded as a V→C dependency — no custom scalar detection pass needed.
  - **InterCoreTransferAndSync**: scalar transfers now use a PIPE_S set/wait pair, isolated from the tensor flag space to avoid deadlocks.
  - **SeparateCVScope**: mechanism A replaces the redundant store→load that the scope clone leaves in VECTOR scopes with the stored value directly (and drops the now-redundant `memref_ext.volatile` annotation).
  - **SSBufferManager**: the memref scalar channel now carries the scalar's own element type through the SSBuffer slot (`memref<T, ssbuf>`), instead of a hardcoded `memref<i32, ssbuf>`. This restores support for f32/f16/i64 scalars that the LLVM→memref migration had dropped (f32 stores used to fail `memref.store` verification).
  - **UT**: `test_scalar_v2c_external_input.mlir` covering the scalar store/load, PIPE_S sync and the typed SSBuffer slot.

### DIFF
--- a/third_party/ascend/include/DynamicCVPipeline/Common/SSBufferManager.h
+++ b/third_party/ascend/include/DynamicCVPipeline/Common/SSBufferManager.h
@@ -107,9 +107,8 @@ getSsbufConstAndPointerCast(OpBuilder &builder, Location loc, uint64_t addr,
       builder.getAttr<hivm::AddressSpaceAttr>(hivm::AddressSpace::SSBUF);
   auto memrefType = MemRefType::get({}, elemType, nullptr, addressSpaceAttr);
 
-  return {addrConst,
-          builder.create<hivm::PointerCastOp>(loc, memrefType,
-                                              addrConst.getResult())};
+  return {addrConst, builder.create<hivm::PointerCastOp>(
+                         loc, memrefType, addrConst.getResult())};
 }
 
 inline hivm::PointerCastOp createPointerCastOp(OpBuilder &builder, Location loc,
--- a/third_party/ascend/include/DynamicCVPipeline/Common/SSBufferManager.h
+++ b/third_party/ascend/include/DynamicCVPipeline/Common/SSBufferManager.h
@@ -98,19 +98,25 @@ inline MemRefType getSsbufMemrefType(Builder &builder) {
 }
 
 inline std::pair<arith::ConstantOp, hivm::PointerCastOp>
-getSsbufConstAndPointerCast(OpBuilder &builder, Location loc, uint64_t addr) {
+getSsbufConstAndPointerCast(OpBuilder &builder, Location loc, uint64_t addr,
+                            Type elemType) {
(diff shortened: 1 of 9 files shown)


third_party/ascend/lib/DynamicCVPipeline/SplitDataflow/SeparateCVScope.cpp
```diff
@@ -1015,6 +1014,54 @@ void mlir::triton::SeparateCVScopePass::runOnOperation() {
...
+        return;
+      }
+      loadOp.replaceAllUsesWith(storeVal);
+      deadLoads.push_back(loadOp);
+    });
+    for (LLVM::LoadOp loadOp : deadLoads) {
+      loadOp->erase();
+    }
+  });
+
```
lrf22, line 1085: 最好提一个函数出来

KanuaK (dismissed):

third_party/ascend/include/DynamicCVPipeline/Common/SSBufferManager.h
```diff
@@ -98,19 +98,24 @@ inline MemRefType getSsbufMemrefType(Builder &builder) {
 }
 
 inline std::pair<arith::ConstantOp, hivm::PointerCastOp>
-getSsbufConstAndPointerCast(OpBuilder &builder, Location loc, uint64_t addr) {
+getSsbufConstAndPointerCast(OpBuilder &builder, Location loc, uint64_t addr,
```
hongziqi, line 101: Please supplement the PR description.

1801ZDL: Done

hongziqi (approved):

WuTYSFG (approved):

PR #1561, opened 2026-08-14
[MaskAnalysis](fix) add atomic op in parseSplat

KanuaK:
## Summary

Fix scalar-splat masks being dropped when the same mask is shared by an indirect load and an atomic RMW operation.

`MaskState::parseSplat` recognizes a splatted `i1` value as a mask only when all of its users consume it as a mask. `LoadOp` and `StoreOp` were handled, but `AtomicRMWOp` was not. Therefore, an atomic user caused the analysis to reject the entire mask, potentially lowering both the load and atomic operation as unconditional single-element accesses.

## Changes

- Recognize `AtomicRMWOp` as a valid mask consumer in `MaskState::parseSplat` to preserve the scalar mask as a dynamic zero-or-one-sized access for both the indirect load and atomic operation.
- Add an end-to-end indirect `atomic_add` test verifying that inactive programs do not update a sentinel expert bucket.


WuTYSFG (approved):

hongziqi (approved):
approve

PR #1618, opened 2026-08-18
[ssbuffer](feat) Analyze the update procees of WhileOp ConditionArgs

sxm-ai:
Analyze the update procees of WhileOp ConditionArgs:
For scf.while with the ssbuffer.main_loop tag, the use-def chain is traversed in reverse order starting from the condition value of scf.condition:
The updates of each round of parameters carried in the loop are traced across the before and after regions. (The before parameter is traced to the operand with the same subscript as the scf.yield in the after region. The after parameter is traced to the argument with the same subscript as the scf.condition forwarding.)
When an op (scf.if or scf.for) in a nested region is encountered, the result is generated by the region terminator, and the operand of the terminator needs to be traced together.
The hit condition is that a tensor value of the tensor type appears in the chain and is generated inside the while loop. Generation inside the loop means that the value is generated by an op in a region of the while loop or is a parameter carried in the while loop or nested loop.
A tensor defined outside the while loop is not a problem. Its non-scalar computation is not within the scope of what needs to be cloned. Only a scalar is extracted from it in the loop.
If the hit condition is met, the fallback attribute is added to the module, and the error code is ERRCODE_IGNORED (indicating that the function is not supported).

# New contributor declaration

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests

### DIFF
--- a/third_party/ascend/include/DynamicCVPipeline/AnalyzeDataFlow.h
+++ b/third_party/ascend/include/DynamicCVPipeline/AnalyzeDataFlow.h
@@ -110,6 +110,27 @@ public:
   }
 };
 
+// Pass for analyzing whether an scf.while condition depends on non-scalar
+// (tensor) data through its loop-carried arg update chain
+class AnalyzeWhileConditionArgsPass
+    : public PassWrapper<AnalyzeWhileConditionArgsPass,
+                         OperationPass<ModuleOp>> {
+public:
+  MLIR_DEFINE_EXPLICIT_INTERNAL_INLINE_TYPE_ID(AnalyzeWhileConditionArgsPass)
+
+  AnalyzeWhileConditionArgsPass() = default;
+
+  void runOnOperation() override;
+
+  llvm::StringRef getArgument() const override {
+    return "analyze-while-condition-args";
+  }
+  llvm::StringRef getDescription() const override {
+    return "Analyze whether scf.while condition args are updated with "
+           "non-scalar data";
+  }
+};
+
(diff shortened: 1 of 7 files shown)


third_party/ascend/lib/DynamicCVPipeline/AnalyzeDataFlow/AnalyzeWhileConditionArgs.cpp
```diff
@@ -0,0 +1,175 @@
...
+                           SmallVectorImpl<Value> &worklist) {
+  Operation *op = result.getOwner();
+
+  if (!whileOp->isProperAncestor(op)) {
+    return;
+  }
+
+  worklist.append(op->operand_begin(), op->operand_end());
+
+  for (Region &region : op->getRegions()) {
```
olivervnc, line 85: ok

olivervnc (approved):

WuTYSFG (approved):

third_party/ascend/include/DynamicCVPipeline/AnalyzeDataFlow.h
```diff
@@ -110,6 +110,27 @@ class AnalyzeScopePass
   }
 };
 
+// Pass for analyzing whether an scf.while condition depends on non-scalar
```
hongziqi, line 113: It is recommended to supplement the PR description.

hongziqi (approved):

PR #1644, opened 2026-08-18
[runtime](fix) evaluate alignment specialization once per launch

Misaka19280hw:
## Summary

- Evaluate the Ascend alignment-specialization policy once per generated binder invocation instead of once per eligible runtime argument.
- Keep the policy evaluation lazy, so binders with no applicable alignment specialization do not pay an extra call.
- Add device-independent regression tests for the one-call and zero-call paths.

## Motivation

This is a performance follow-up to #1480.

The generated binder currently embeds `backend.use_alignment_specialization(options)` in every eligible argument expression. All arguments in one binder invocation share the same backend and raw options, so these calls return the same value. A kernel with `N` eligible runtime arguments therefore performs the same Python method call `N` times on every launch.

The supplied softmax reproducer has four eligible runtime arguments and showed an approximately 1 microsecond host-launch regression per kernel. The optimization computes the policy once in the generated function and reuses that invocation-local value. It is recomputed for every launch, so switching compile options on the same binder remains correct.

This does not change specialization keys, compilation behavior, generated kernels, or the number of kernel launches.

## Performance

Measured on Ascend 910B4 with the supplied softmax kernel, using the same compiled kernel and cache for both binder implementations:

| Metric | Current | Optimized |
|---|---:|---:|
| Alignment-policy calls per launch | 4 | 1 |
| Median host enqueue time | 60.430 us/kernel | 59.286 us/kernel |

- Median paired improvement: `0.842 us/kernel`.
- Bootstrap 95% confidence interval: `[0.530, 1.412] us/kernel`.
- Samples: 201 paired samples, 5 launches per sample.
- Compilation events during measurement: 0.
- Numerical correctness: PASS.
- The five requested kernel launches remain five kernel launches.

An independent cross-process binder benchmark measured `4.236 us` before and `3.301 us` after, an improvement of approximately `0.935 us/kernel`.

## Testing

- Wheel build: PASS.
- CPU/native specialization regression: `163 passed`.
- Ascend 910B4 specialization integration with a fresh Triton cache: `24 passed`.
- Ascend 910B4 repository softmax smoke: `2 passed`; maximum difference from PyTorch was `3.0517578125e-05`.
- Release patch replay: `git apply --check` PASS.
- `git diff --check`: PASS.
- Applicable pre-commit checks, including YAPF, Ruff, AST, mypy, whitespace, EOF, merge-conflict, and private-key checks: PASS.

After rebasing onto the latest `main`, the CPU/native regression was rerun: `163 passed in 0.71s`.

### DIFF
--- a/third_party/ascend/patch/triton-ascend-3.6.0.patch
+++ b/third_party/ascend/patch/triton-ascend-3.6.0.patch
@@ -1118,13 +1118,15 @@ index cd871cb2e..82eb6960d 100644
                  if triton.knobs.compilation.front_end_debugging:
                      raise
 diff --git a/python/triton/runtime/jit.py b/python/triton/runtime/jit.py
-index 0f506818f..b2992457f 100644
+index 0f506818f..43742c328 100644
 --- a/python/triton/runtime/jit.py
 +++ b/python/triton/runtime/jit.py
-@@ -398,14 +398,16 @@ def create_function_from_signature(sig, kparams, backend):
+@@ -398,14 +398,19 @@ def create_function_from_signature(sig, kparams, backend):
      assert len(sig.parameters) == len(kparams)
      # Create the function argument list and the dict entries for the return statement
      specialization = []
++    uses_alignment_specialization = False
++    alignment_specialization_name = "__triton_alignment_specialization"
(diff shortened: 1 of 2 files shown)


Misaka19280hw:
Closing this draft because it targeted `main`. The same patch will be reopened against the development branch `main-dev`.

PR #1651, opened 2026-08-19
[launch](feat) optimize the launch to reduce the exection time

LH-123L:
# Launcher Performance Optimization and Code Combination Design Document #

## Overview ##

Optimize the code structure and content of the launch template to ensure that the code is elegant and maintainable and the performance is improved.

### 1.1 Motivation ###

`make_launcher`Reconstructs a large number of static C++ strings in the function body each time it is called. (device pointer auxiliary function, msprof extern declaration, msprof callback, pre-launch timing code, alignment auxiliary function, tensor shape extraction function, and so on) These strings do not depend on any kernel signature or metadata and are completely unchanged during the process life cycle. However, they are repeatedly spliced into f-strings, resulting in unnecessary string allocation and formatting overhead.

### 1.2 Change Content ###

Move the following 6 static C++ code segments from the`make_launcher`The function body is upgraded to a module-level constant.

 *  `_CPP_DEVICE_POINTER`\:`DevicePtrInfo`Structure and`getPointer()`Auxiliary functions (including interned functions)`"data_ptr"`key optimization)
 *  `_CPP_MSPROF_EXTERN`\: extern "C" declaration of msprof, flag variable,`ProfCtrlHandle`Callback
 *  `_CPP_MSPROF_CALLBACK`\: During module initialization`MsprofRegisterCallback`Invoking
 *  `_CPP_MSPROF_BEFORE_LAUNCH`\: sampling of the cycle time before launch
 *  `_CPP_ALIGN_LAUNCH_OFFSET`\: Inline function for parameter offset alignment.
 *  `_CPP_GET_TENSOR_SHAPE`\: auxiliary function for extracting a shape from the PyObject tensor.

Change the function body to direct reference:

```
cpp_device_pointer = _CPP_DEVICE_POINTER
cpp_msprof_extern = _CPP_MSPROF_EXTERN
cpp_msprof_callback = _CPP_MSPROF_CALLBACK
cpp_msprof_call_before_launch = _CPP_MSPROF_BEFORE_LAUNCH
```

### 1.3 Dead Code Deletion ###

Remove 4 variables that were never referenced by C++ templates after calculation:

 *  `launch_signature_items`
 *  `launch_arg_count`
 *  `launch_arg_ptrs`
 *  `launch_arg_sizes`

### 1.4 Combining NPUUtils Single Instances ###

In the original code`make_launcher`Internally called twice`NPUUtils()`\: Obtains the number of cores at one time, and reads the number of cores at the other time.`npu_utils_mod.__file__`. Due to the`NPUUtils.__new__`A singleton is implemented, and the second instantiation does not recompile.`.so`, but it will still be repeated`__init__`The importlib spec creation and module loading in are combined into one.`npu_utils = NPUUtils()`Directly reused after invoking.

--------------------

### 2.1 Zero copy kernel parameter transfer ###

**Before the optimization:**`triton_launch_kernel`Put each`kernel_args[arg_idx]`Copy to a separate`std::vector<char>`(`copied_kernel_args`), and then after packing to`launch_args`Perform memcpy again. Each parameter generates one heap allocation + two memcpy.

**After optimization: deleted**`copied_kernel_args`, directly from the`kernel_args[arg_idx]`memcpy to the final thread_local continuous buffer.

**Security:**`triton_async_launch()`(See`npu_utils.cpp`) Synchronous execution on the caller thread`launch_call`lambda (via`OpCommand::Run`), so the caller provides the`kernel_args`/`arg_sizes`The pointer is valid throughout the lambda life cycle. This assumption is explicitly recorded in the comments in the generated code.

### 2.2 Multiplexing of the thread_local launch parameter buffer ###

**Before optimization: A new is constructed each time the launch is launched.**`std::vector<char> launch_args(total_size, 0)`to trigger heap allocation and initialization.

**After optimization:**

```
static thread_local std::vector<char> launch_args;
launch_args.assign(total_size, 0);
```

`assign()`Memory is reused when the existing capacity is sufficient, and is reallocated only when the parameter size increases.`triton_async_launch`Synchronous execution. The thread_local does not have cross-thread life cycle issues.

### 2.3 Replace METHOD_VARARGS with METHOD_FASTCALL. ###

Python launcher extension module`launch`Entrance from`METH_VARARGS`Switch to`METH_FASTCALL`\:

 *  Avoids the overhead of constructing temporary tuples for location parameters at the Python call layer.
 *  Manually Parse Parameters (Direct Index)`args[i]`), skip`PyArg_ParseTuple`Format String Interpreter for
 *  Pointer parameter usage`reinterpret_cast<...>(PyLong_AsUnsignedLongLong(...))`, borrow reference semantics and`"O"`Consistent format

### 2.4 Interned dict key ###

`packedMetadata`Dictionary lookup from`PyDict_GetItemString`(Each call constructs a temporary`PyUnicode`) Use the static interned string in the function instead:

```
static PyObject* key_kernel_name = PyUnicode_InternFromString("kernel_name");
PyObject* kernelNameObj = PyDict_GetItemWithError(packedMetadata, key_kernel_name);
```

`tensor_kinds`The key is processed in the same way. Also, use`PyDict_GetItemWithError`In addition, the KeyError is set when the key is missing, which fixes the silently ignored error in the original implementation.

### 2.5 FFTS Address Query Encapsulation ###

The original code directly invokes the deprecated`aclrtGetHardwareSyncAddr`. After the optimization, the operation is passed.`get_backend_func("get_ffts_addr", ...)`Indirect invoking, and obtaining the FFTS address is unified as follows:

```
void* ffts_addr = nullptr;
ret = get_ffts_addr(stream, &ffts_addr);
```

This allows the FFTS implementation to replace at the backend adaptation layer, while allowing extension points for subsequent caches of FFTS addresses (to avoid system calls every launch).

### 2.6 Returning an Empty Grid in Advance ###

Empty grid protection is added to the two launch paths.

```
if (gridX <= 0 || gridY <= 0 || gridZ <= 0) {
  printf("WARNING: Skipping launch ...\n");
  return;
}
```

Do not trigger meaningless workspace allocation and kernel invoking when the grid is compressed to 0 by coalescing or other compile-time transformations.

### 2.7 Other Improvements ###

 *  Use C++11`nullptr`Replace`NULL`
 *  Uniform pointer declaration style (`Type* ptr`Instead of`Type *ptr`)
 *  Uniform brace indentation style
 *  `PyModuleDef`Used by sentinel`nullptr`
 *  This interface is invoked during module initialization.`init_npu_utils()`Moved dlsym queries such as g_async_launch from the per-call path to one-time initialization.
 *  The return value after the hook function is invoked is correct.`Py_DECREF`(Original code leaks hook return value reference)
 *  `tensorKinds`Parsing time`Py_DECREF(kind)`Fix sequence item reference leaks

--------------------

### 3.1 Analysis of Duplicate Codes ###

The two launch functions differ in the parameter packaging mode.

| Aspects                  | triton_launch_kernel (C ABI)                                             | _launch (local Python)                                                               |
| ------------------------ | ------------------------------------------------------------------------ | ------------------------------------------------------------------------------------ |
| Parameter Source         | `const void* const* kernel_args`\+`arg_sizes`Runtime array               | Named C++ parameters known at compile time`arg0, arg1, ...`                          |
| Parameter Package        | thread_local`std::vector<char>`\+ Runtime alignment calculation + memcpy | `struct __attribute__((packed))`Compilation-time layout + aggregation initialization |
| Kernel launch parameters | `launch_args.data(), launch_args.size()`                                 | `&args, sizeof(args)`                                                                |
| Application Scenario     | External ABI stable path, supporting the number of dynamic parameters    | Python launcher directly invokes the call, with zero overhead.                       |

The logic of the two functions is the same before the parameter packaging and after the kernel launch function is invoked, including:

1.  workspace pointer declaration, grid coalescing,`blockNum4Workspace`, pre_launch (first time), workspace memory allocation, and RAII guard
2.  start lambda, pre_launch (second time),`blockNum`Calculation, grid threshold-crossing warning printing, auto-map block cap,`mixBlockNumRation`/`nodeBasicBlockDim`1. Enable the device print tunnel, query the FFTS address, and allocate and initialize the sync block lock.
3.  Timing before and after msprof, device print tunnel disabling, stream synchronization/taskqueue return, and async_launch invoking

### 3.2 Three-Segment Sharing Template Design ###

In the`make_launcher`Extract three f-string templates internally:

#### `_launch_preamble` ####

Located on top of the function body, outside the lambda:

```
_launch_preamble = f"""
  void* workspace_addr_ptr = nullptr;
  void* workspace_handle = nullptr;
  {coalesce_grid_div}
  uint32_t blockNum4Workspace = gridX * gridY * gridZ;
  {get_backend_func("pre_launch", True)}
  {workspace 分配块（条件）}"""
```

#### `_launch_lambda_pre` ####

At the beginning of lambda, before the parameter packing:

```
_launch_lambda_pre = f"""  {lambda 头 if taskqueue} {{
    {get_backend_func("pre_launch", False)}
    uint32_t blockNum = gridX * gridY * gridZ;
    {grid 警告 #ifdef block}
    {auto-map cap 条件}
    uint32_t mixBlockNumRation = {mix_block_dim_ratio};
    uint32_t nodeBasicBlockDim = (mixBlockNumRation << 16) + blockNum;
    {device print Open 条件}
    aclError ret = ACL_SUCCESS;
    {FFTS 查询 条件}
    {sync block lock 分配块（条件）}
    {workspace ret 检查 条件}"""
```

#### `_launch_lambda_post` ####

After the parameter package, end lambda:

```
_launch_lambda_post = f"""
    {cpp_msprof_call_before_launch}
    __KERNEL_LAUNCH_CALL__
    {device print Close 条件}
    {cpp_msprof_call_after_launch}
    {return ret if taskqueue else aclrtSynchronizeStream}
  }};
  {async_launch 调用 if taskqueue}
  return;
}}"""
```

### 3.3 Differentiation Injection ###

`__KERNEL_LAUNCH_CALL__`Placeholder passes through the final f-string assembly`.replace()`Inject the unique kernel launch call of each function.

```
#triton_launch_kernel
{_launch_lambda_post.replace('__KERNEL_LAUNCH_CALL__', cpp_kernel_launch)}

#_launch Path
{_launch_lambda_post.replace('__KERNEL_LAUNCH_CALL__', cpp_kernel_launch_local)}
```

The unique logic of the two functions remains in place:

 *  `triton_launch_kernel`Tensor shape/kinds parsing`launch_arg_sizes`Construction,`reserve_slot`lambda, thread_local buffer allocation and memcpy
 *  `_launch`\: Packed struct definition and aggregation initialization

### 3.4 Kernel launch Invoking and Constructing Auxiliary Functions ###

Later, further`cpp_kernel_launch`It's different from the`cpp_kernel_launch_local`The construct of is merged into an internal helper function:

```
def _make_kernel_launch(args_ptr, args_size, indent="    "):
    cfg = "&cfgCfgInfo" if (compile_on_910_95 and enable_simt) else "nullptr"
    cfg_setup = ...  #91095 SIMT Dynamic UBUF Configuration (Conditional)
    return f"""{cfg_setup}{indent}ret = aclrtLaunchKernelWithHostArgs(
        func, blockNum, stream, {cfg}, {args_ptr}, {args_size}, nullptr, 0);
"""
```

By parameterizing`args_ptr`(`launch_args.data()`vs`&args`),`args_size`(`launch_args.size()`vs`sizeof(args)`) and indent level, eliminating the common/SIMT configuration x two paths in a total of four segments duplicated.`aclrtLaunchKernelWithHostArgs`Invoke the template.

### 3.5 Changes in the Code Quantity ###

 *  Eliminate about 70 duplicate lines of C++ code templates (Python f-string level)
 *  The generated C++ source code behavior remains equivalent. The logic, such as workspace, sync lock, msprof, and grid coalescing, appear once in each of the two paths. The consistency is ensured by the shared template.

--------------------

## 4. Test Coverage ##

Test File`test_launcher_export_api.py`Statically generated code validation is provided for the above optimizations. (No NPU hardware or compiled C extensions required to run):

| Test Case                                                           | Verification Content                                                                                                                |
| ------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------- |
| `test_make_launcher_exposes_triton_launch_kernel`                   | C ABI export signature; confirm None`copied_kernel_args`;Check that the thread_local buffer exists.                                 |
| `test_make_launcher_resolves_npu_utils_from_active_cache_root`      | Source code generation does not depend on the cache_root absolute path (cache key certainty).                                       |
| `test_make_launcher_shrinks_coalesced_grid_for_both_launch_paths`   | The grid coalescing takes effect in both paths.                                                                                     |
| `test_make_launcher_uses_ceil_div_for_row_coalescing`               | In ceil-div mode, coalescing takes effect on both paths.                                                                            |
| `test_make_launcher_enables_91095_simt_for_sls_mixed_parallel_mode` | In 910_95 SIMT mode, cfgCfgInfo is generated correctly in both paths, and the parameter pointer is correct.                         |
| `test_make_launcher_block_cap_uses_only_env_and_blacklist`          | The auto-map block cap function is controlled only by E && !B and is irrelevant to row coalescing. Assert the two paths separately. |
| `test_merged_code_workspace_allocation_appears_in_both_paths`       | The workspace allocation and RAII guard appear once in each path.                                                                   |
| `test_merged_code_sync_block_lock_appears_in_both_paths`            | The sync block lock allocation and guard appear once in both paths.                                                                 |
| `test_merged_code_msprof_calls_in_both_paths`                       | msprof timers occur once in each path                                                                                               |
| `test_merged_code_preamble_shared_variables_present`                | Shared variables such as workspace pointer, blockNum4Workspace, blockNum, and ret are declared in both paths.                       |
| `test_merged_code_taskqueue_mode_in_both_paths`                     | The launch_call/std::function/async_launch in taskqueue mode exists in both paths.                                                  |
| `test_merged_code_grid_warning_in_both_paths`                       | `ENABLE_GRID_WARN_PRINT`Overrun warnings controlled by environment variables exist in both paths                                    |
| `test_argument_packing_differs_between_launch_paths`                | The parameter packing modes of the two paths are unique (thread_local memcpy vs packed struct).                                     |
| `test_npu_launcher_exposes_launcher_so_path`                        | The NPULauncher correctly caches the launcher file and exposes the launcher .so path.                                               |

auxiliary function`_split_launch_functions(src)`Press the`static void _launch(`Divide the generated source code into C ABI segments and local segments, so that each test can assert two paths separately, instead of depending on the global.`src.count()`.

--------------------

### 5.1 Why Two Sets of Parameter Packing Mechanisms Are Retained ###

`triton_launch_kernel`The C ABI path of the requires the number and type of parameters to be determined dynamically at run time (passed in by an external caller).`(void**)`Pointer array), which must be continuous buffer + memcpy.`_launch`The path is directly invoked by the Python extension module in the same compilation unit. The parameter type is known in the codegen. The packed struct can be used with zero extra copy and zero runtime alignment calculation. The two paths are applicable to different scenarios. Forcible unification will compromise the performance of one path.

### 5.2 Why use string .replace() instead of nested f-string ###

`__KERNEL_LAUNCH_CALL__`The placeholder scheme makes`_launch_lambda_post`It can be constructed as a complete f-string just once, and differentiated content injected through simple string substitution. Using nested f-strings or additional format calls increases the cognitive complexity and the brace escape burden of the template. This placeholder is used only in the codegen phase and does not appear in the final generated C++ source code.

### 5.3 Why Not Caching Environment Variables ###

`TRITON_GRID_WARN_PRINT`,`TRITON_ENABLE_TASKQUEUE`,`TRITON_DEVICE_PRINT`Wait for environment variables to be in the`make_launcher`internal read. Although it is theoretically possible to cache to avoid duplicate`os.getenv`, however, these variables may be dynamically modified by users or tests at run time (especially in device print debugging scenarios), and caching can destroy the dynamic switching capability.`make_launcher`It is protected by the Triton compilation cache (the same kernel signature is codegen only once), and repeated.`getenv`The overhead of is negligible.

### 5.4 Zero-copy thread-safe boundary ###

The security of zero copy and thread_local buffer depends on`triton_async_launch`The implementation details of launch_call are executed synchronously. This constraint is explicitly recorded in two places: (1) generated C++ code comments; (2) This document, 2.1. If the future`npu_utils.cpp`Medium`triton_async_launch`To perform lambda asynchronously instead, you need to revert the parameter copy or introduce a synchronous primitive.

--------------------

## 6. List of documents ##

| File                                                                | Change Type                                                                |
| ------------------------------------------------------------------- | -------------------------------------------------------------------------- |
| `third_party/ascend/backend/driver.py`                              | Core Optimization and Code Consolidation                                   |
| `third_party/ascend/backend/backend_register.py`                    | Concomitant submission`4fb2eb73`Small adaptation of                        |
| `third_party/ascend/unittest/pytest_ut/test_launcher_export_api.py` | Test the clearance, deduplication, and new combination verification cases. |

## 7. Optimization effect ##

Take the softmax operator as an example. The /driver.py(202): call (this track) takes about 30% less time.

|                         | Maximum value | Minimum value | Mean   | Median |
| ----------------------- | ------------- | ------------- | ------ | ------ |
| Before the optimization | 139.400       | 62.770        | 78.326 | 65.85  |
| After optimization      | 76.830        | 42.05         | 48.34  | 45.35  |

### DIFF
--- a/third_party/ascend/backend/backend_register.py
+++ b/third_party/ascend/backend/backend_register.py
@@ -355,4 +355,4 @@ def async_launch(func):
      fprintf(stderr, "Error: triton_async_launch is unavailable\\n");
      return;
    }}
-   g_async_launch(static_cast<void*>(&{func}), name.c_str());'''
+   g_async_launch(static_cast<void*>(&{func}), kernelName);'''
--- a/third_party/ascend/backend/driver.py
+++ b/third_party/ascend/backend/driver.py
@@ -672,17 +672,17 @@ def make_launcher(constants, signature, metadata):
          lockOffset += syncBlockLockStrideI64) {{
       lockInitData[lockOffset] = syncBlockLockParticipantNum;
     }}
-    ret = rtMemcpy(syncBlockLock_ptr, syncBlockLockSize,
+    ret = aclrtMemcpy(syncBlockLock_ptr, syncBlockLockSize,
                    reinterpret_cast<void *>(lockInitData.data()),
-                   syncBlockLockSize, RT_MEMCPY_HOST_TO_DEVICE);"""
+                   syncBlockLockSize, ACL_MEMCPY_HOST_TO_DEVICE);"""
     elif lock_init_value == 0:
-        lock_init_stmt = ("ret = rtMemsetAsync(syncBlockLock_ptr, syncBlockLockSize, 0, "
+        lock_init_stmt = ("ret = aclrtMemsetAsync(syncBlockLock_ptr, syncBlockLockSize, 0, "
                           "syncBlockLockSize, stream);")
     else:
         lock_init_stmt = (f"std::vector<int64_t> lockInitData({lock_num}, {lock_init_value});\n"
-                          "    ret = rtMemcpy(syncBlockLock_ptr, syncBlockLockSize, "
+                          "    ret = aclrtMemcpy(syncBlockLock_ptr, syncBlockLockSize, "
                           "reinterpret_cast<void *>(lockInitData.data()), syncBlockLockSize, "
-                          "RT_MEMCPY_HOST_TO_DEVICE);")
+                          "ACL_MEMCPY_HOST_TO_DEVICE);")
     bs_task_type = metadata.bs_task_type if hasattr(metadata, 'bs_task_type') else 0
     mix_mode = metadata.mix_mode
     compile_on_910_95 = metadata.compile_on_910_95
@@ -1212,13 +1212,14 @@ void triton_launch_kernel(const char* kernelName, aclrtFuncHandle func, aclrtStr
     return;
   }}
   std::vector<size_t> launch_arg_sizes;
-  launch_arg_sizes.assign(arg_sizes, arg_sizes + num_args);
-  // NOTE: We deliberately do not copy kernel_args into a separate buffer here.
-  // triton_async_launch() invokes the launch_call synchronously on the caller
-  // thread (see npu_utils.cpp triton_async_launch -> OpCommand::Run), so the
-  // caller-provided kernel_args/arg_sizes memory remains valid for the entire
-  // lifetime of launch_call. Avoiding the per-arg vector<vector<char>> copy
-  // removes num_args heap allocations per launch.
+  launch_arg_sizes.reserve(num_args);
+  std::vector<std::vector<char>> copied_kernel_args;
+  copied_kernel_args.reserve(num_args);
+  for (int arg_idx = 0; arg_idx < num_args; ++arg_idx) {{
+    launch_arg_sizes.push_back(arg_sizes[arg_idx]);
+    copied_kernel_args.emplace_back(arg_sizes[arg_idx]);
+    memcpy(copied_kernel_args.back().data(), kernel_args[arg_idx], arg_sizes[arg_idx]);
+  }}
 
   // Only 1D parallelization is supported for NPU.
   // Pointer type becomes flattened 1-D Memref tuple: base_ptr, data_ptr,
@@ -1252,8 +1253,7 @@ void triton_launch_kernel(const char* kernelName, aclrtFuncHandle func, aclrtStr
     {'size_t dtdata_offset = reserve_slot(sizeof(void*), 8);' if enable_device_print else ''}
     size_t total_size = args_offset;
 
-    static thread_local std::vector<char> launch_args;
-    launch_args.assign(total_size, 0);
+    std::vector<char> launch_args(total_size, 0);
     {'memcpy(launch_args.data() + ffts_offset, &ffts_addr, sizeof(void*));' if target_support_ffts else ''}
     {f'memcpy(launch_args.data() + sync_block_lock_offset, &syncBlockLock_ptr, sizeof(void*));' if not metadata.force_simt_only else ''}
     {f'memcpy(launch_args.data() + workspace_offset, &workspace_addr_ptr, sizeof(void*));' if not metadata.force_simt_only else ''}
@@ -1261,7 +1261,7 @@ void triton_launch_kernel(const char* kernelName, aclrtFuncHandle func, aclrtStr
     for (int arg_idx = 0; arg_idx < num_args; ++arg_idx) {{
       size_t alignment = launch_arg_sizes[arg_idx] >= 8 ? 8 : (launch_arg_sizes[arg_idx] >= 4 ? 4 : 1);
       kernel_arg_offset = _align_launch_offset(kernel_arg_offset, alignment);
-      memcpy(launch_args.data() + kernel_arg_offset, kernel_args[arg_idx], launch_arg_sizes[arg_idx]);
+      memcpy(launch_args.data() + kernel_arg_offset, copied_kernel_args[arg_idx].data(), launch_arg_sizes[arg_idx]);
       kernel_arg_offset += launch_arg_sizes[arg_idx];
     }}
     memcpy(launch_args.data() + grid_offset, &gridX, sizeof(int32_t));
--- a/third_party/ascend/backend/driver.py
+++ b/third_party/ascend/backend/driver.py
@@ -426,6 +426,7 @@ def generate_npu_header_src():
 #include <assert.h>
 #include <stdbool.h>
 #include <cstdlib>
+#include <cstring>
 #include <string>
 #include <memory>
 #include <sys/syscall.h>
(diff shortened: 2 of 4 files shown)


third_party/ascend/backend/backend_register.py
```diff
@@ -355,4 +355,4 @@ def async_launch(func):
      fprintf(stderr, "Error: triton_async_launch is unavailable\\n");
      return;
    }}
-   g_async_launch(static_cast<void*>(&{func}), name.c_str());'''
+   g_async_launch(static_cast<void*>(&{func}), kernelName);'''
```
hongziqi, line 358: please fix conflicts first

LH-123L:
#1645
增加了profile_scratch和global_scratch参数，

hongziqi (approved):

KanuaK (approved):

PR #1653, opened 2026-08-19
[Autotune](fix) extract packed_metadata from tuple when using @libentry

gymgit1:
Background:
Autotuner._make_kernel_call attempts to extract packed_metadata from the return value of self.fn.run() to get the kernel_name.

Root Cause:
When a kernel is decorated with @libentry(), the call chain becomes Autotuner -> LibEntry -> JITFunction. To pass additional info like constexprs, the LibEntry.run() method wraps its return value into a tuple: (CompiledKernel, constexprs).
The old code directly called getattr(res, "packed_metadata", None) on this tuple. Since tuples do not have this attribute, it returned None, causing the Autotuner to fail to capture the target kernel name.

Solution:
Add a type check before accessing the attribute. If res is a tuple, extract the first element (the CompiledKernel object) before fetching packed_metadata. This approach safely handles all return types: tuples from @libentry, raw CompiledKernel objects, and FutureKernel during async compilation.

# New contributor declaration

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests

### DIFF
--- a/third_party/ascend/backend/runtime/autotuner.py
+++ b/third_party/ascend/backend/runtime/autotuner.py
@@ -2265,6 +2265,8 @@ class AutoTilingTuner(Autotuner):
                     *args,
                     **current,
                 )
+                if isinstance(res, tuple):
+                    res = res[0]
                 packed_metadata = getattr(res, "packed_metadata", None)
                 if isinstance(packed_metadata, dict):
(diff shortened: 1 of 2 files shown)


zhudada0120 (commented):
Please consider adding a regression test.

KanuaK (approved):

hongziqi (approved):

PR #1682, opened 2026-08-20
[CI](feat) migrate image, wheel and LLVM builds to self-hosted runners

xuedinge233:
### Summary

Migrate the three build workflows off GitHub-hosted runners with local Docker daemon, onto the self-hosted buildkit runners (`linux-amd64-cpu-4 / linux-aarch64-cpu-4`) using `docker/build-push-action@v7` against the remote buildkitd. All mirrors now go through the internal cache service instead of public registries/mirrors.

### Changes

Release Triton Ascend Images (build-docker-image.yml)

- Re-enabled `push/pull_request` triggers and added concurrency control.
- Build per-architecture via matrix (`amd64/arm64`), each job runs on its own buildkit runner inside the CANN container; removed `setup-qemu/setup-buildx`.
- `docker/build-push-action@v6` → `@v7;` tags get a `-amd64/-arm64` suffix so the two jobs never race on the same tag.
- Layer cache moved from `type=gha` (unusable with remote buildkitd) to `type=registry` on SWR, keyed by CANN version + arch.
- All apt/pip/PyTorch downloads redirected to the internal `cache-service` mirrors.
- Fixed dubious ownership git error in the container via `safe.directory`.

Docker image (docker/Dockerfile)

- Mirror URLs parameterized as build args (`APTMIRROR, PIP_INDEX_URL, PIP_TRUSTED_HOST, TORCH_INDEX_URL`); base image now pulls from internal SWR.
- apt update/install wrapped in a 3-attempt retry loop with proper exit-on-final-failure (the old single attempt died on short read mid-transfer).
- `torch==2.7.1+cpu` is a transitive dep of `torch-npu` and only exists on the PyTorch CPU index — installed separately via `--find-links` with `--timeout 600`, then the rest of the requirements install with `--timeout 300`.

Wheels (wheels.yml + wheels.Dockerfile)

- Replaced `cibuildwheel` (requires docker run, unavailable on these runners) with `docker/build-push-action@v7`; wheels are extracted with outputs:` type=local` instead of `docker cp`.
- New `wheels.Dockerfile` builds inside the manylinux image, detects the setup script location per branch (`setup_ascend.py` → `setup.py` → `python/setup.py`), and exports only the .whl files via a scratch stage.
- MAX_JOBS capped at 8 (nproc / 6) to avoid OOM on memory-constrained workers.
- Removed host-side `setup-python/cache` steps — no Python needed on the runner anymore.

LLVM Build (llvm-build.yml + almalinux.Dockerfile))

- AlmaLinux path: switched from `docker build + docker create + docker cp` (needs Docker daemon) to `docker/build-push-action@v7` with target: output and outputs: type=local; the `Dockerfile packages /install and /sccache` as tarballs in a scratch output stage; base image parameterized to the internal SWR mirror.
- Ubuntu path: fixed `mlir-opt` link being `OOM-killed` on 4-CPU pods by installing lld-15 (so -`DCMAKE_LINKER=lld` actually takes effect) and adding `-DLLVM_PARALLEL_LINK_JOBS=1`.
- Removed ninja `check-mlir` from the Ubuntu build — 3500+ upstream tests were failing on 2 Bytecode cases and the test run isn't part of the artifact.
- Both paths use the internal `cache-service` for apt/pip.

### Verification

- `x86_64` image build and wheel build pass on the new runners; AlmaLinux LLVM build passes via buildkitd.
- `aarch64` wheel build and Ubuntu LLVM build are tracked for memory-related stability (buildkitd connection reset issues are infra-side and being investigated separately).

### DIFF
--- a/.github/workflows/build/almalinux.Dockerfile
+++ b/.github/workflows/build/almalinux.Dockerfile
@@ -1,5 +1,6 @@
 # https://github.com/AlmaLinux/container-images/blob/9f9b3c8c8cf4a57fd42f362570ff47c75788031f/default/amd64/Dockerfile
-FROM almalinux:8.10-20250411
+ARG BASE_IMAGE=almalinux:8.10-20250411
+FROM ${BASE_IMAGE}
 ARG llvm_dir=llvm-project
 # Add the cache artifacts and the LLVM source tree to the container
 ADD sccache /sccache
@@ -10,6 +11,9 @@ ENV SCCACHE_CACHE_SIZE="2G"
 RUN dnf install --assumeyes llvm-toolset
 RUN dnf install --assumeyes python39-pip python39-devel git
 RUN alternatives --set python3 /usr/bin/python3.9
+
+ENV PIP_INDEX_URL=http://cache-service.nginx-pypi-cache.svc.cluster.local/pypi/simple \
+    PIP_TRUSTED_HOST=cache-service.nginx-pypi-cache.svc.cluster.local
 
 RUN python3 -m pip install --upgrade pip
 RUN python3 -m pip install --upgrade cmake ninja sccache lit nanobind
@@ -23,8 +27,6 @@ RUN cmake -GNinja -Bbuild \
   -DCMAKE_C_COMPILER=clang \
   -DCMAKE_CXX_COMPILER=clang++ \
   -DCMAKE_ASM_COMPILER=clang \
-  -DCMAKE_C_COMPILER_LAUNCHER=sccache \
-  -DCMAKE_CXX_COMPILER_LAUNCHER=sccache \
   -DCMAKE_CXX_FLAGS="-Wno-everything" \
   -DCMAKE_LINKER=lld \
   -DCMAKE_INSTALL_PREFIX="/install" \
@@ -44,4 +46,13 @@ RUN cmake -GNinja -Bbuild \
 
 RUN ninja -C build install
 
-RUN ninja check-mlir -C build
+# ---------------------------------------------------------------------------
+# Package build outputs for extraction via --output type=local (no need for
+# docker cp / Docker daemon on the runner).
+# ---------------------------------------------------------------------------
+RUN tar czf /llvm.tar.gz -C / install \
+    && tar czf /sccache.tar.gz -C / sccache
+
+FROM scratch AS output
+COPY --from=0 /llvm.tar.gz /
+COPY --from=0 /sccache.tar.gz /
--- a/.github/workflows/llvm-build.yml
+++ b/.github/workflows/llvm-build.yml
@@ -3,18 +3,18 @@ name: LLVM Build
 # TEMPORARILY DISABLED — all triggers commented out so this workflow never
 # starts and never occupies a runner. Uncomment to re-enable.
 on:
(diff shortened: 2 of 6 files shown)


.github/workflows/build/almalinux.Dockerfile
```diff
@@ -1,5 +1,6 @@
 # https://github.com/AlmaLinux/container-images/blob/9f9b3c8c8cf4a57fd42f362570ff47c75788031f/default/amd64/Dockerfile
-FROM almalinux:8.10-20250411
+ARG BASE_IMAGE=almalinux:8.10-20250411
```
hongziqi, line 2: Why are there 43 workflows for this PR? Could you clarify which ones will rarely run going forward? Additionally, do we need to rebuild the images when making LLVM‑related changes?

xuedinge233: Because this involves modifying the entire image and building related processes, each of them will have multiple tasks. Each task has a corresponding trigger path, and most of the time they will not be invoked.

WuTYSFG (dismissed):

.github/workflows/llvm-build.yml
```diff
@@ -1,22 +1,24 @@
...
-  # pull_request:
-  #   paths:
-  #     - cmake/llvm-hash.txt
-  #     - third_party/ascend/patch/**
-  #     - .github/workflows/llvm-build.yml
-  #     - .github/workflows/llvm-build/**
+  push:
+    paths:
+      - cmake/llvm-hash.txt
+      - third_party/ascend/patch/**
```
hongziqi: Specify llvm_patch

.github/workflows/llvm-build.yml
```diff
@@ -1,22 +1,24 @@
...
+      - .github/workflows/llvm-build.yml
+      - .github/workflows/build/almalinux.Dockerfile
+  pull_request:
+    paths:
+      - cmake/llvm-hash.txt
+      - third_party/ascend/patch/**
```
hongziqi: Specify llvm_patch

.github/workflows/wheels.yml
```diff
@@ -1,17 +1,21 @@
...
+  pull_request:
+    paths:
+      - .github/workflows/wheels.yml
+      - .github/workflows/build/wheels.Dockerfile
+      - setup.py
+      - pyproject.toml
+      - version.txt
+      - cmake/llvm-hash.txt
+      - cmake/json-version.txt
+      - third_party/ascend/patch/**
```
hongziqi: remove this line

.github/workflows/llvm-build.yml
xuedinge233: Modified.

.github/workflows/llvm-build.yml
xuedinge233: Modified.

.github/workflows/wheels.yml
xuedinge233: Modified.

hongziqi (approved):

WuTYSFG (approved):

insanecoder-zzj (approved):

PR #1699, opened 2026-08-20
[Launcher](fix) defer NPU utils initialization before fork

zhanwei33:
## Summary

- Port the fork-safety fix from #974/#982 to the current `main-dev`.
- Defer `npu_utils.so` compilation and loading until an actual NPU utility consumer needs it.
- Avoid importing `torch_npu` while selecting the backend or resolving build include/library paths.
- Serialize concurrent `npu_utils.so` cache builds.
- Load `npu_utils.so` from the exact path resolved by Triton's cache manager while preserving relocatable launcher behavior.

## Background

#974 fixed the issue on `release/3.2.2-dev`, and #982 applied the same fix to `release/3.2.2`. The equivalent change was not present in `main-dev`.

Before this change, constructing the Ascend driver also constructed `NPUUtils`, which could:

1. import `torch_npu` while selecting the backend or calculating build parameters;
2. compile `npu_utils.so`;
3. load the extension into the current Python process.

Importing TorchNPU or loading the extension may initialize TorchNPU/CANN runtime state. If the process forks afterwards, the child can inherit runtime threads, locks, and device state that are not safe to reuse after `fork()`, potentially causing the child process to hang.

This change removes driver construction and build-path discovery as eager TorchNPU/CANN initialization points.

### Lazy NPU utility initialization

`NPUUtils.__init__()` now initializes Python-side fields only.

- `get_so_path()` builds or retrieves `npu_utils.so` on demand.
- `_load_mod()` loads the Python extension only when a Python-side NPU utility operation needs it.
- Repeated construction of the singleton no longer rebuilds or reloads the extension.

### Avoid eager `torch_npu` imports

Torch and TorchNPU package directories are resolved with
`importlib.util.find_spec()` instead of importing the packages.

This allows backend selection and compiler flag construction to complete without triggering TorchNPU runtime initialization.

### Serialize cache construction

Concurrent processes may request the same `npu_utils.so` cache entry.

The build path now:

1. checks the cache;
2. acquires the cache build lock;
3. checks the cache again;
4. builds and publishes the shared object only if it is still missing.

This avoids concurrent writes and redundant builds.

### Use the cache-manager-resolved shared-object path

The generated launcher exposes `set_npu_utils_path()`.

After importing the launcher extension, `NPULauncher` passes the exact result of:

```python
NPUUtils().get_so_path()

### DIFF
--- a/third_party/ascend/backend/backend_register.py
+++ b/third_party/ascend/backend/backend_register.py
@@ -18,6 +18,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 
+import importlib.util
 import os
 from typing import Callable, Dict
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: third_party/ascend/backend/backend_register.py [hunk 1]
```diff
@@ -195,12 +196,21 @@ def get_cc_cmd():
     ]
 
 
+def _get_package_dir(package_name):
+    spec = importlib.util.find_spec(package_name)
```
zhanwei33: 检视意见：

这里使用 `importlib.util.find_spec()` 的目的是只查询 Python 模块的加载信息，而不为了获取安装路径直接执行顶层包的 `__init__.py`。例如：

```python
spec = importlib.util.find_spec("torch_npu")
```

可能得到：

```text
spec.origin = "/path/to/site-packages/torch_npu/__init__.py"
spec.submodule_search_locations = ["/path/to/site-packages/torch_npu"]
```

对于 package，应优先从 `submodule_search_locations` 取得包目录；对于单文件 module，则回退到 `origin` 所在目录。这样可以拼出 `torch/include`、`torch_npu/include` 和 `torch_npu/lib`，同时避免仅为了读取 `__file__` 而直接执行 `import torch_npu`，降低在 fork 前触发 TorchNPU/CANN 运行时初始化的风险。

zhanwei33:
@

PR #1700, opened 2026-08-20
[compiler](NFC) consolidate Ascend compile options

zhanwei33:
## Summary

- Remove or internalize the audited Ascend compile options while preserving backend-derived behavior.
- Preserve `limit_auto_multi_buffer_buffer` and costmodel controls, and rename cache-count metadata to `buf_slot_num_of_*`.
- Rebase the series onto current `main-dev` and reconcile the upstream deprecated-option compatibility boundary.

## Validation

- `pytest -q third_party/ascend/unittest/pytest_ut/test_compiler_layout_memory_contract.py`
- `pytest -q third_party/ascend/unittest/autotune_ut/test_autotune_param_valid.py::test_grid_num_tiles_is_not_auto_injected`
- `pytest -q third_party/ascend/unittest/pytest_ut/test_layout_memory_895_closure_differential.py::test_grid_num_tiles_is_not_auto_injected`
- `pytest -q third_party/ascend/unittest/pytest_ut/test_driver.py::test_legacy_msprof_register_env_does_not_skip_launcher`
- `python -m unittest discover -s third_party/ascend/unittest/costmodel_ut -p "test_costmodel*.py" -v`
- `pytest -q third_party/ascend/unittest/costmodel_ut/test_compiler_costmodel_contract.py`

All validation above is source/unit-level; no NPU kernel was launched.

PR #1701, opened 2026-08-20
[Ascend](fix) complete global scratch runtime setup

Ritsuka314:
## Why

PR #1645 added compiler-emitted global scratch metadata and launch support. Once a kernel requests nonzero scratch, an ordinary Ascend launch still fails with "kernel requires a runtime memory allocation, but no allocator was set" unless the application explicitly configures Triton's generic allocator. The driver also reports a placeholder max_shared_mem value of one byte, so resource validation can reject kernels that use on-chip UB.

The integration suite also leaks its NPU default device from the unified-attention test into later files on the same pytest-xdist worker. Depending on worker scheduling, the pointer-bitcast CPU references are then created on NPU and fail with a CPU/NPU device mismatch.

## What

- Report 248 KiB of usable UB for Ascend910_95 and Ascend950 targets, and 192 KiB for other supported Ascend targets.
- When Triton's allocator is unset, allocate global scratch through the active Torch-NPU or MindSpore backend; an explicit triton.set_allocator callback still takes precedence.
- Allocate scratch as bytes, overallocate as needed, and expose a compiler-aligned pointer. Adapt MindSpore's _data_ptr interface to Triton's Buffer protocol.
- Retain the scratch owner through deferred task-queue callbacks so its device pointer remains valid until the ACL launch is enqueued.
- Keep NPU_DEVICE_LIMIT core-count handling unchanged.
- Scope the unified-attention test's NPU default device to each test case so the xdist worker state is restored afterward.
- Add host coverage for architecture limits, fallback allocation, explicit allocator precedence, alignment/MindSpore adaptation, and queued-launch ownership.

## Tests

- Launcher/driver host regression: 52 passed.
- Compiler layout/memory contract: 3 passed, 30 environment-specific skips.
- pre-commit on all changed files: passed.
- The default-device leak was reproduced in PR #1701 and independently in PRs #1689 and #1700 plus a main-dev push; all showed the same four pointer-bitcast device-mismatch failures.
- No global-scratch NPU device test was run on this machine; task-queue enabled/disabled device validation is handed off separately.

## Checklist

- [x] I have written a PR description following the project rules.
- [x] I have run pre-commit over the PR range.
- [x] I have added tests.
- [x] I have not added any lit tests.

### DIFF
--- a/third_party/ascend/backend/backend_register.py
+++ b/third_party/ascend/backend/backend_register.py
@@ -75,6 +75,20 @@ class _LazyBackendStrategyRegister:
 backend_strategy_registry = _LazyBackendStrategyRegister()
 
 
+class _ScratchBuffer:
+
+    def __init__(self, tensor, alignment):
+        self.tensor = tensor
+        get_data_ptr = getattr(tensor, "data_ptr", None)
+        if get_data_ptr is None:
+            get_data_ptr = tensor._data_ptr
+        address = get_data_ptr()
+        self._aligned_data_ptr = ((address + alignment - 1) // alignment) * alignment
+
+    def data_ptr(self):
+        return self._aligned_data_ptr
+
+
 @backend_strategy_registry.register("mindspore", "version_hash")
 def version_hash():
     import mindspore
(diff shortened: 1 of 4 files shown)


Ritsuka314:
Closing because this change needs to be handled with the existing global-scratch work.

Ritsuka314:
Reopening to carry the max_shared_mem follow-up for the merged global-scratch work.

hongziqi (dismissed):

WuTYSFG (dismissed):

PR #1704, opened 2026-08-21
[Costmodel](fix) remove compile options and env variable

BloodOrange666:
## Background

This PR simplifies the costmodel integration in `triton-ascend`. It removes the dedicated costmodel compile option and the costmodel-specific worker-count environment variable, so the costmodel runtime becomes a regular prediction interface that can be scheduled by upper layers.

## Main Changes

1. Removed the `enable_costmodel_backend` compile option.
   - `NPUOptions` no longer exposes this field.
   - `parse_options` no longer changes `use_bytecode` based on a costmodel-specific option.
   - The costmodel examples no longer pass `enable_costmodel_backend`.

2. Removed the `TRITON_COSTMODEL_WORKER_NUM` environment variable.
   - `costmodel_runtime.py` no longer owns an internal thread pool.
   - Costmodel parallelism is delegated to the caller, such as Inductor precompile parallelism or the autotuner's existing scheduling logic.

3. Changed `costmodel_bench` to accept a single item.
   - The new API evaluates one dict item per call and returns `(config, latency_us)`.
   - Callers with multiple configs can call `costmodel_bench(item)` per item and aggregate the returned pairs themselves.
   - This makes each config item the scheduling unit and allows finer-grained parallel execution in the upper layer.

4. Fixed disk cache loading for costmodel latency.
   - `get_cache_manager().get_file()` returns the cache file path, not the JSON payload.
   - The implementation now opens the returned path and reads the payload with `json.load()`.
   - The in-process memory cache is still kept so repeated predictions in the same process can be reused.

5. Updated tests and documentation.
   - Costmodel runtime tests now cover single-item input, failure fallback, and memory/disk cache reuse.
   - The compiler contract test now verifies that `use_bytecode` remains a normal compile option.
   - Chinese and English examples now show single-item `costmodel_bench` calls and caller-side aggregation.


KanuaK (dismissed):

docs/en/examples/09_costmodel_example.md
```diff
@@ -3,8 +3,8 @@
 This example shows the basic costmodel backend flow:
 
 - generate TTIR from a Triton frontend kernel;
-- build `costmodel_bench` inputs for multiple candidate configs;
```
KanuaK, line 6: Please supplememt PR desciption.

KanuaK (approved):

WuTYSFG (approved):

PR #1737, opened 2026-08-24
[simd_simt](fix) use getRealReductionOps in ScanConverter

zhangzhengshi:
## Why

`ScanConverter::convertToTargetOp` used `getReductionOps`, which returns every op in the scan body (except the terminator).

When `TRITON_DEBUG=1`, the frontend inserts overflow-check assert semantics (`tt.assert` plus its condition-computing ops) directly into the combine body alongside the real reduction op. With the naive `getReductionOps`, `reductionOps.front()` could end up being one of these inserted assert-related ops instead of the actual add/mul/max/min op, which breaks `isReductionOpSupported` whenever `TRITON_DEBUG=1` is set on a `tl.cumsum`/`cumprod`/`cummax`/`cummin`.

## What

Switch `getReductionOps` to `getRealReductionOps`, which computes a backward slice from the body's yielded value so only ops that actually feed the scan result are kept (it also already filters `arith.extf`/`arith.truncf`/`arith.bitcast` precision casts).

## Testing

- `TRITON_DEBUG=1` + `tl.cumsum` now correctly dispatch to the right cumsum library function.

---
# New contributor declaration

  - [x] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests
  - [] This PR does not need a test because `FILL THIS IN`.

  - [] I have not added any `lit` tests.

### DIFF
--- a/third_party/ascend/lib/TritonToLinalg/TritonOpConverter.cpp
+++ b/third_party/ascend/lib/TritonToLinalg/TritonOpConverter.cpp
@@ -1264,7 +1264,7 @@ LogicalResult
 ScanConverter::convertToTargetOp(triton::ScanOp op,
                                  typename triton::ScanOp::Adaptor adaptor,
                                  ConversionPatternRewriter &rewriter) const {
-  auto reductionOps = this->getReductionOps(op);
+  auto reductionOps = this->getRealReductionOps(op);
   if (reductionOps.empty()) {
     return rewriter.notifyMatchFailure(op,
                                        "No reduction op found in scan body");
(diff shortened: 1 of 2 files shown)


third_party/ascend/lib/TritonToLinalg/TritonOpConverter.cpp
```diff
@@ -1264,7 +1264,7 @@ LogicalResult
 ScanConverter::convertToTargetOp(triton::ScanOp op,
                                  typename triton::ScanOp::Adaptor adaptor,
                                  ConversionPatternRewriter &rewriter) const {
-  auto reductionOps = this->getReductionOps(op);
+  auto reductionOps = this->getRealReductionOps(op);
```
hongziqi, line 1267: add test cases

WuTYSFG (approved):

hongziqi (approved):

PR #1755, opened 2026-08-25
[compiler](fix) restore enable_select_analysis option

zhanwei33:
## Summary

- restore `enable_select_analysis` as a public `NPUOptions` field with its legacy default of `True`
- forward the configured value to the Triton-to-Linalg lowering pass instead of forcing it to `True`
- remove the option from the deprecated-and-ignored compatibility list

## Motivation

PR #1700 removed the public field and fixed select analysis to `True`. That changes the established compile-option contract because callers can no longer disable select analysis. Restore the original configurable behavior while retaining the other compile-option cleanup.

## Validation

- `pre-commit run --files third_party/ascend/backend/compiler.py third_party/ascend/backend/utils.py`
- `git diff --check`

No unit tests are added by this PR.


zhanwei33:
Closing this PR because main-dev should keep select analysis fixed to True. The remaining requirement is compatibility: legacy enable_select_analysis inputs must be accepted with a deprecation warning and ignored without an error or assertion.

PR #1788, opened 2026-08-27
[Launcher](fix) prebuild NPU utils without torch_npu loading

zhanwei33:
## Summary

- Build or retrieve `npu_utils.so` when the `NPUUtils` singleton is first initialized.
- Keep `NPUUtils` construction limited to building/caching the shared object; do not load it as part of `NPUUtils.__init__()`.
- Load `npu_utils.so` during `NPULauncher` initialization through `set_npu_utils_path()`, before the first kernel launch.
- Preserve the #1699 behavior that resolves Torch and TorchNPU package paths without importing `torch_npu`.
- Restore the original cache-miss build flow without adding a cross-process file lock.
- Build only once for repeated singleton construction and retain retry behavior when the initial build fails.

## Background

This is an alternative initialization strategy based on #1699.

#1699 defers both construction and loading of `npu_utils.so`. This PR separates the phases at a different boundary: `NPUDriver` construction prebuilds and caches the shared object, while `NPULauncher` initialization loads it and resolves the required symbols. Kernel launch therefore does not pay first-use `dlopen()` overhead.

Because cache-key generation and compiler-path discovery no longer import `torch_npu`, compiling and linking the wrapper does not execute TorchNPU/CANN initialization in the Python process.

## Behavior

On first `NPUUtils()` construction:

1. initialize `_cache_path` and `npu_utils_mod`;
2. check the existing `npu_utils.so` cache entry;
3. build and publish the shared object directly when the cache misses;
4. retain the resolved path while leaving `npu_utils_mod` as `None`;
5. mark the singleton initialized only after the build succeeds.

Loading remains separate from construction:

- `NPUUtils.load_binary()` loads the Python extension through `_load_mod()` when binary loading requires it.
- `NPULauncher.__init__()` imports the launcher and calls `set_npu_utils_path()` with the resolved path.
- `set_npu_utils_path()` performs `dlopen()` and resolves the launcher helper symbols before the first kernel launch.

## Validation

- `25 passed`:
  - `third_party/ascend/unittest/pytest_ut/test_backend_npu_utils_build_flags.py`
  - `third_party/ascend/unittest/pytest_ut/test_launcher_export_api.py`
- Python syntax checks passed.
- Configured pre-commit hooks passed for the changed files.
- Real NPU runtime validation has not yet been run for this alternative initialization timing.

## Trade-offs

- Creating `NPUDriver` now requires the compiler, CANN headers/libraries, and TorchNPU development files even when no kernel is launched.
- Separate processes that miss the same cache entry concurrently may perform redundant builds; this keeps the build flow consistent with the implementation before #1699.

This PR is intended to evaluate these trade-offs against fully lazy construction.

### DIFF
--- a/third_party/ascend/backend/backend_register.py
+++ b/third_party/ascend/backend/backend_register.py
@@ -18,6 +18,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 
+import importlib.util
 import os
 from typing import Callable, Dict
 
@@ -195,12 +196,21 @@ def get_cc_cmd():
     ]
 
 
+def _get_package_dir(package_name):
+    spec = importlib.util.find_spec(package_name)
+    if spec is None:
+        raise ModuleNotFoundError(f"No module named '{package_name}'")
+    if spec.submodule_search_locations:
+        return os.path.realpath(next(iter(spec.submodule_search_locations)))
+    if spec.origin is None:
+        raise ModuleNotFoundError(f"Cannot locate module '{package_name}'")
+    return os.path.dirname(os.path.realpath(spec.origin))
+
+
 @backend_strategy_registry.register("torch_npu", "get_cc_cmd_npu_utils")
 def get_cc_cmd_npu_utils():
-    import torch
-    import torch_npu
-    torch_path = os.path.dirname(os.path.realpath(torch.__file__))
-    torch_npu_path = os.path.dirname(os.path.realpath(torch_npu.__file__))
+    torch_path = _get_package_dir("torch")
+    torch_npu_path = _get_package_dir("torch_npu")
     cc_cmd = [
         f"-I{os.path.join(torch_path, 'include')}",
         f"-I{os.path.join(torch_npu_path, 'include')}",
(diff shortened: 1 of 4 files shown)


third_party/ascend/backend/backend_register.py
```diff
@@ -195,12 +183,21 @@ def get_cc_cmd():
     ]
 
 
+def _get_package_dir(package_name):
+    spec = importlib.util.find_spec(package_name)
```
zhanwei33, line 187: 这里使用 `importlib.util.find_spec()` 的目的是只查询 Python 模块的加载信息，而不为了获取安装路径直接执行顶层包的 `__init__.py`。例如：

```python
spec = importlib.util.find_spec("torch_npu")
```

可能得到：

```text
spec.origin = "/path/to/site-packages/torch_npu/__init__.py"
spec.submodule_search_locations = ["/path/to/site-packages/torch_npu"]
```

hongziqi (dismissed):

KanuaK (dismissed):

zhanwei33:
retry

hongziqi (approved):

WuTYSFG (approved):

KanuaK (approved):

PR #1838, opened 2026-08-28
[TritonToLinalg](fix) boundary size Error for make_block_ptr

liwei-link:
## Description
When lowering tt.make_block_ptr with boundary_check, getBoundarySizes() decomposes the flattened linear offset back into per-axis offsets by dividing/moding it with each axis stride. Two issues made this crash on valid inputs:

Zero-stride axis: If a boundary-checked axis has stride 0, the decomposition divides by zero. Such an axis carries no address contribution, so its boundary cannot be (and need not be) reconstructed from the offset — but the code still tried to.
Wrong-operand conversion in arithmetic helpers: In add/sub/mul/div/rem/minOpFoldResult, the branch handling a dynamic (non-constant) RHS mistakenly applied convertToIndexIfNeeded to lhsValue and asserted on lhs. Whenever a non-index-typed Value appeared as the RHS, the assert fired instead of the conversion happening.

## Changes
getBoundarySizes():
Skip boundary reconstruction on zero-stride checked axes with a warning and keep the current block size for that axis, instead of dividing by stride 0.
Only fold offsetShift % stride into the running remainder when the stride is non-zero.
Fix the copy-paste bug in addOpFoldResult / subOpFoldResult / mulOpFoldResult / divOpFoldResult / remOpFoldResult / minOpFoldResult: convert the RHS (not LHS) to index type in the dynamic-RHS branch.
remOpFoldResult: fold x % 1 to 0.

# New contributor declaration

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests

### DIFF
--- a/third_party/ascend/lib/Utils/Utils.cpp
+++ b/third_party/ascend/lib/Utils/Utils.cpp
@@ -400,22 +400,29 @@ getBoundarySizes(llvm::ArrayRef<int32_t> boundaryCheck, Value ptr,
       curPtrOffset, fullShapeReCast.getConstifiedMixedOffset(), loc, rewriter);
 
   for (int i = 0; i < shapedType.getRank(); ++i) {
+    OpFoldResult curStride = fullShapeReCast.getConstifiedMixedStrides()[i];
     if (llvm::find(boundaryCheck, i) != boundaryCheck.end()) {
-      auto fullShape = fullShapeReCast.getConstifiedMixedSizes()[i];
+      if (isZero(curStride)) {
+        emitWarning(loc)
+            << "getBoundarySizes() cannot reconstruct boundary on checked "
+               "zero-stride axis "
+            << i << "; keep current block size for this axis";
+        continue;
+      }
 
-      OpFoldResult curOffset = divOpFoldResult(
-          offsetShift, fullShapeReCast.getConstifiedMixedStrides()[i], loc,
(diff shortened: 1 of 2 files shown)


third_party/ascend/lib/Utils/Utils.cpp
```diff
@@ -400,22 +400,29 @@ getBoundarySizes(llvm::ArrayRef<int32_t> boundaryCheck, Value ptr,
       curPtrOffset, fullShapeReCast.getConstifiedMixedOffset(), loc, rewriter);
 
   for (int i = 0; i < shapedType.getRank(); ++i) {
+    OpFoldResult curStride = fullShapeReCast.getConstifiedMixedStrides()[i];
```
hongziqi, line 403: It is recommended to supplement the PR description.

hongziqi, line 403: It is also recommended to supplement relevant test cases.

liwei-link: supplied

liwei-link: the relevant test cases will be added in a separate PR.

WuTYSFG (approved):

hongziqi (approved):

PR #1912, opened 2026-09-02
[Autotune](feat) Add JIT-scoped cache for autotune compile failures

CHNJZ:
## Summary

Add an in-memory negative cache to the Ascend `AutoTilingTuner` for Configs that fail during compilation.

When the autotune key changes, autotuning is performed again. Successfully compiled Configs may reuse the JIT kernel cache, but failed Configs previously had no corresponding cache entry and were compiled again.

This change skips a previously failed Config when its actual JIT compilation inputs remain unchanged.

## Implementation

- Generate the failure-cache key from the JIT compilation identity, including:
  - JIT function source and dependencies
  - target and device
  - JIT specialization and compiler options
  - cache-invalidating environment variables
  - existing UBTuner compiler options
- Keep the autotune key out of the failure-cache key so failures can be reused across different tuning keys.
- Retry compilation when constexpr values, compiler options, target, source, or other JIT inputs change.
- Cache the following per-Config failures:
  - `CompileTimeAssertionFailure`
  - `MLIRCompilationError`
  - `OutOfResources`
- Do not cache the original failure when UBTuner successfully recovers the Config.
- Disable failure caching conservatively when user-defined or JIT pre-run hooks may modify the compilation inputs.
- Store only the exception type to avoid retaining and repeatedly printing large compiler error messages or IR dumps.

The cache is local to the current `AutoTilingTuner` instance and is not persisted to disk.

## Validation

Added a vector-add reproducer containing:

- An invalid Config that triggers a real UB overflow:
  - `BLOCK_SIZE=20480`
  - `ITERATIONS=8`
- A valid Config:
  - `BLOCK_SIZE=1024`
  - `ITERATIONS=1`

The test runs the kernel using the key sequence `K1 -> K2 -> K1`:

1. `K1` compiles the invalid Config once and records the UB-overflow failure.
2. `K2` causes an autotune cache miss but skips the invalid Config through the new failure cache.
3. The final `K1` invocation hits the normal autotune memory cache.

The observed compile-failure counts are:

```text
[1, 0, 0]
```

The output is also compared with the PyTorch reference to verify correctness.
image
image

### DIFF
--- a/third_party/ascend/backend/runtime/autotuner.py
+++ b/third_party/ascend/backend/runtime/autotuner.py
@@ -30,6 +30,7 @@ import gc
 import inspect
 import os
 import pprint
+import threading
 import time
 import warnings
 from concurrent.futures import ThreadPoolExecutor
@@ -41,7 +42,10 @@ from typing import Any, Dict, List, Optional, Tuple
 from torch import Tensor
 
 import triton
+from triton import knobs
+from triton._C.libtriton import get_cache_invalidating_env_vars
 from triton.runtime.autotuner import Autotuner, Config
+from triton.runtime.jit import compute_cache_key
 from triton.backends.ascend.utils import (_InternalNPUOptionInt, _remove_deprecated_npu_options,
                                           _RESERVED_NPU_OPTION_NAMES, is_compile_on_910_95)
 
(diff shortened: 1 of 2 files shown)


third_party/ascend/backend/runtime/autotuner.py
```diff
@@ -2191,11 +2203,126 @@ def _print_benchmark_results(self, timings) -> None:
...
+        if config.pre_hook is not None or self.user_defined_pre_hook:
+            return None
+
+        jit_fn = self.fn
+        visited = set()
+        while not isinstance(jit_fn, triton.runtime.JITFunction):
+            if jit_fn is None or id(jit_fn) in visited:
+                return None
+            visited.add(id(jit_fn))
+            jit_fn = getattr(jit_fn, "fn", None)
```
HinPeng: Unwrapping heuristics bypasses the constexpr values it injects, so failures may be cached under the wrong compile key?

CHNJZ: The implementation has been updated so compile-cache-key generation now follows the same effective argument-processing order as the actual kernel invocation:

_make_kernel_call → Heuristics.run → JITFunction.run → binder

third_party/ascend/backend/runtime/autotuner.py
HinPeng, line 1: Add some uts.

CHNJZ: done

KanuaK (approved):

hongziqi (approved):

PR #2026, opened 2026-09-07
[Atomic](feat) Fold identity-valued selects into atomic masks

happyhappyhappy-arch:
Representing inactive elements with `tl.where` and an identity value, such as zero for atomic add, can therefore still issue unnecessary GM accesses and incur substantial memory traffic and contention.

This PR recovers access predicates from atomic values. For example:

```python
atomic_add(ptr, where(active, value, 0), mask)
```

becomes:

```python
atomic_add(ptr, value, mask & active)
```

Predicate recovery follows nested selects and supported shape operations, including splat, broadcast, expand-dims, reshape, and transpose. It reuses the RMW identity table and handles operation-specific identities, including floating MIN/MAX infinities and all-ones values for AND and UMIN. Unused-result atomics whose input is entirely an identity value are eliminated. Shared value users are preserved, and recovery does not cross arithmetic or bitcasts.

The optimization is implemented in `TritonToGraph/LegacyMemoryAccess` and registered as a normal `GraphOptimizationRule` in the existing `graph-optimize` pass. It is enabled by default, requires no graph analyses, and follows the standard rule-mask and rewrite-budget controls. 

Elimination requires an unused atomic result and targets numerical reductions whose identity updates are not synchronization events and whose exact signed-zero, NaN, and denormal effects need not be preserved. Surviving atomics retain their operation, memory semantics, and scope.


third_party/ascend/backend/compiler.py
```diff
@@ -185,6 +185,9 @@ def make_ttir(mod, metadata, opt):
             ub_capacity_bytes=graph_ub_budget_bytes_for_arch(opt.target_arch),
             compile_mode=opt.compile_mode,
         )
+    if opt.compile_mode == "simt_only":
```
KanuaK: Need to discuss if it is suitable to add pass to simt-only compile mode.

third_party/ascend/lib/DiscreteMaskAccessConversion/DiscreteMaskAccessConversionPass.cpp
```diff
@@ -449,38 +449,19 @@ struct DiscreteMaskAtomicConversion
...
-        {RMWOp::XOR, TypelessValue::Zero},
-        {RMWOp::XCHG, TypelessValue::Undefined},
-    };
-    assert(initMap.find(rmwOp) != initMap.end());
-    auto typelessVal = initMap.at(rmwOp);
-    if (typelessVal == TypelessValue::Undefined) {
-      // Undefined default value atomic op will be decomposed in AscendNPU-IR
-      op->setAttr(ConverterUtils::discreteMaskAttrName,
-                  UnitAttr::get(rewriter.getContext()));
+    auto attr = getAtomicRMWIdentityAttr(rmwOp, src.getType(), rewriter);
```
kaixin1976: why need to modify this file ?

third_party/ascend/lib/TritonToGraph/LegacyMemoryAccess/AtomicMaskCanonicalization.cpp
```diff
@@ -0,0 +1,256 @@
...
+  Attribute attr = constant.getValue();
+  if (auto dense = dyn_cast<DenseElementsAttr>(attr)) {
+    if (!dense.isSplat())
+      return false;
+    attr = dense.getSplatValue<Attribute>();
+  }
+  // Compare with the selected RMW's identity. Numeric FP equality handles
+  // +/-infinity for MIN/MAX and both signs of zero for FADD uniformly.
+  if (auto fp = dyn_cast<FloatAttr>(attr))
+    if (auto expected = dyn_cast<FloatAttr>(identity))
```
kaixin1976: 1. why put this opt pass under a folder called "legacyxxxx" ?
2.  this is not a canonization, this is graph optimization  .
3.

third_party/ascend/lib/TritonToGraph/GraphOptimizationPass.cpp
```diff
@@ -62,9 +62,10 @@ namespace triton {
 namespace cfg {
 namespace {
 
-constexpr std::array<GraphOptimizationRulePhase, 11> kRulePhases = {
+constexpr std::array<GraphOptimizationRulePhase, 12> kRulePhases = {
     GraphOptimizationRulePhase::DiagonalMaskRemoval,
     GraphOptimizationRulePhase::ConvertModuloToMask,
+    GraphOptimizationRulePhase::AtomicMaskCanonicalization,
```
Popeyesxs, line 68: here is a question. the kRulePhases  will have other pass now. By your pass, if there is 64 atomic operators, other pass will lose effects.

third_party/ascend/lib/TritonToGraph/LegacyMemoryAccess/AtomicMaskCanonicalization.cpp
```diff
@@ -0,0 +1,256 @@
...
+      predicate.getLoc(), rewriter.getOneAttr(predicate.getType()));
+  return rewriter.create<arith::XOrIOp>(predicate.getLoc(), predicate, one);
+}
+
+// Follow only select/shape definitions of the atomic value. For example,
+// select(a, identity, select(b, v, identity)) yields (v, !a & b).
+// Do not traverse arithmetic, loads or casts: their results need not preserve
+// the identity (e.g. 0 * NaN). Other users of the selects remain untouched.
+// Return (value, mask); a null mask means no identity branch was found.
+static std::pair<Value, Value> recoverAtomicContribution(Value value,
```
Popeyesxs, line 112: The pass will generate a mask. But the mask may have confilce with Discrecte mask pass

third_party/ascend/lib/TritonToGraph/LegacyMemoryAccess/AtomicMaskCanonicalization.cpp
```diff
@@ -0,0 +1,256 @@
...
+  }
+
+  LogicalResult apply(IRRewriter &rewriter) override {
+    auto identity = matchAtomicMask(op);
+    if (failed(identity))
+      return failure();
+    rewriter.setInsertionPoint(op);
+    auto src = op.getVal();
+
+    if (isAtomicIdentity(src, *identity)) {
```
Popeyesxs, line 191: When we wanna cancel the identity atomic, I think that we should check the memory seqlence first. acq_re or sys scope, I think that the atomic cannot be cancelled.

PR #2040, opened 2026-09-08
[TritonToLinalg](fix) preserve mixed if result types

elstainniles:
## Summary

Preserve non-pointer results when lowering mixed-result `scf.if` operations, and fix the handling of loop-carried boolean masks.

The changes preserve the existing structured-access path for a restricted class of prefix-mask recurrences, route other runtime loop masks through supported masked-access lowering, and report an explicit compilation error when that lowering is unavailable.

This PR does not implement general multi-element dynamic loop-mask support on A3.

## Motivation

Two separate issues need to be addressed:

1. **Premature conversion of non-pointer `scf.if` results.**

   The scalar-pointer carrier conversion also applied the global type converter to sibling results. This could turn a `tensor<i1>` mask into a memref while its producers and consumers still expected tensor semantics, causing conversion/materialization failures.

2. **Incorrect treatment of loop-carried masks as loop-invariant.**

   Mask analysis followed a loop argument back to its initial value without checking whether the backedge updated it. For a recurrence such as `mask = mask & current`, this can lose the intersection with conditions from previous iterations.

   For example, with prefix limits changing from `4` to `12`, the correct active range remains the first four lanes. Replacing the current mask with its initial value can incorrectly reactivate lanes 4–11.

   A related liveness issue could erase mask-producing operations while the loop still required their results, producing an unresolved materialization to `tensor<16xi1>`.

Previously, depending on the input and lowering path, these cases could fail compilation, compile with incorrect results, or compile and execute correctly under narrower conditions.

### 1. Preserve mixed `scf.if` result types

- Keep the existing integer-address transport for scalar pointer results.
- Preserve the original types of non-pointer sibling results, including tensor masks.
- Do not use the pointer-carrier rewrite as a general tensor-to-memref conversion.
- Add the GM address-space annotation to remaining scalar-pointer carriers that are consumed directly, before removing their pass-local provenance marker.

Tensor masks remain tensors at this control-flow boundary. Consumers that explicitly require buffers, and subsequent bufferization, retain responsibility for their memory representation. The global boolean-storage conversion policy is unchanged.

### 2. Canonicalize supported prefix-mask recurrences

Use the identity:

    (lane < bound) & (lane < limit) == lane < min(bound, limit)

Replace the loop-carried boolean tensor with an exact scalar boundary, allowing the existing range analysis and `memref.subview` / `memref.copy` lowering to remain applicable.

The rewrite is intentionally restricted to:

- An ordinary `scf.for` with exactly one ranked tensor<i1> loop-carried slot.
- A one-dimensional static shape.
- The same fixed, zero-based `tt.make_range`, defined outside the loop.
- Signed `<` comparisons against a splatted scalar or dense-splat bound.
- A direct AND update between the current loop mask and the comparison, in either operand order.
- An initial prefix mask, or an all-true/all-false initial mask.

Loops marked as CFO pointer-descriptor boundaries are excluded.

The rewrite preserves pre-update uses, post-update uses, other loop-carried values, final mask results, and zero-trip behavior. Limits do not need to be monotonic: `4 -> 12` correctly becomes an accumulated boundary of `4 -> 4`.

### 3. Require proof before following a loop mask to its initializer

For loop-carried ranked tensor<i1> values:

- `scf.for`: follow the initializer only when the corresponding yield forwards the same iter argument unchanged.
- `scf.while`: verify the actual condition/yield slot mapping, including permuted before/after slots.
- Return analysis failure when invariance cannot be established.

This prevents an evolving mask from being analyzed as its initial value.

### 4. Route unresolved runtime loop masks explicitly

For relevant load/store/atomic-RMW operations, masks that depend on loop state and cannot be parsed by mask analysis receive runtime-mask markers.

The consumer then uses:

- **Supported singleton load/store:** extract a scalar predicate and guard the actual memory operation with `scf.if`.
- **Supported A5 template path:** preserve the current mask through masked indirect lowering.
- **Otherwise:** emit an explicit unsupported diagnostic rather than silently using the initial mask or an unguarded full-load-and-select fallback.

The A5 indirect load/store path requires `simd_simt_template`, a supported scalar base plus tensor offsets, and rank <= 5. Atomic operations have additional operation/type restrictions.

An unsuccessful runtime-mask conversion does not continue into every legacy fallback path; this is a compatibility limitation of the current implementation.

## Before/after support comparison

This table describes load/store operations in the TritonToLinalg pipeline. It does not describe the separate `simt_only` pipeline.

“Rejected” below means that the mask remains a multi-element runtime loop mask after preceding analysis and canonicalization.

| Pattern | Before the loop-mask changes | A5: SIMD | A5: template | A3: SIMD or template |
| --- | --- | --- | --- | --- |
| Provably invariant, analyzable range mask | Existing range path could handle it | Retained | Retained | Retained |
| Supported prefix-AND recurrence in `scf.for` | Could fail compilation or lose mask history; some cases produced correct results | Scalar boundary + structured access | Scalar boundary + structured access | Scalar boundary + structured access |
| Supported singleton dynamic load/store mask | Input/path-dependent; no dedicated runtime-loop-mask branch | Guarded scalar access | Guarded scalar access | Guarded scalar access |
| Multi-element suffix or discrete-bitmap recurrence outside the prefix rewrite | Input/path-dependent; both working and failing cases existed | Rejected | Masked indirect, subject to interface restrictions | Rejected |
| Dynamic multi-element tensor-mask recurrence retained in `scf.while` | No sound general invariant-mask assumption | Rejected | Masked indirect, subject to interface restrictions | Rejected |

The prefix rewrite does not cover arbitrary multidimensional regions, multiple mask carriers, OR/XOR recurrences, or arbitrary nested control flow. Some such inputs may use the A5 indirect path, but they are not comprehensively validated by this PR.

### Compatibility caveat

The new rejection is a tightening of accepted inputs, not proof that every previously accepted kernel was incorrect.

For example, a legacy full-load-and-select path can produce correct values when all accessed addresses are valid, the input is ordinary read-only memory without concurrent writes, and the current mask is preserved for value selection. Some such inputs may now be rejected on A3 or A5 SIMD.

A5 template inputs that do not satisfy the indirect interface may also lose access to a legacy fallback. A complete before/after compatibility audit has not been performed.

The four A5-only runtime-mask pytest cases now skip on non-A5 targets. This adjusts test expectations; it does not add A3 support or prove the absence of compatibility regressions.

## Performance considerations

- **Invariant/range masks:** retain the existing access path where analysis succeeds.
- **Supported prefix recurrences:** retain structured `subview + copy` lowering and carry history in a scalar boundary instead of forcing per-lane indirect access. Boundary calculations and any required mask reconstruction still have a cost.
- **Singleton masks:** use conditional scalar accesses.
- **Other supported A5 runtime masks:** use masked indirect lowering, with potential additional mask computation, address-generation, and template-call overhead. Dense contiguous accesses may be slower than structured copies; sparse accesses may avoid unnecessary memory traffic.

Generated IR has been inspected to confirm that the tested prefix cases retain `arith.minsi`, `memref.subview`, and `memref.copy`.

No dedicated latency or throughput benchmark has been completed. This PR therefore does not claim zero performance regression or a measured speedup.

Earlier numerical runs included the existing SSBUFFER fallback, which can disable the dynamic CV pipeline. Those results do not establish correctness or performance with that optimization forcibly enabled.

## Regression coverage

- Mixed scalar-pointer/tensor-mask `scf.if` results, covering both branches.
- Prefix recurrence history, including shrink-then-expand limits.
- Zero iterations, negative and oversized bounds, narrow/empty initial masks.
- Pre-update mask uses, final mask outputs, and commuted AND operands.
- Invariant and evolving `scf.while` masks with permuted slots.
- A5 masked-indirect lowering and explicit unsupported-mode diagnostics.
- Singleton accesses guarded by the current mask.
- Masked stores preserving untouched sentinel values.

The numerical references retain the original boolean recurrence and use `rtol=0, atol=0`.

The `isin_by_search_kernel.mlir` expectations are updated to check the actual loop carriers and the current-mask extraction followed by a guarded load.


third_party/ascend/lib/TritonToUnstructure/UnstructureConversionPass.cpp
```diff
@@ -753,6 +754,67 @@ LogicalResult UnstructuredMemAccessConverter<MemAccOpTy>::matchAndRewrite(
...
+  // Handle these accesses before any speculative fallback IR is created.
+  // A failed match must leave the original operation intact for the final
+  // unsupported-mask diagnostic, without repeatedly inserting dead helpers.
+  if (runtimeLoopMask) {
+    if constexpr (std::is_same_v<MemAccOpTy, triton::LoadOp> ||
+                  std::is_same_v<MemAccOpTy, triton::StoreOp>) {
+      // A one-element tensor needs only a scalar predicate, including in SIMD
+      // mode. Use explicit control flow: the legacy scalar masked-load
+      // converter otherwise loads unconditionally and selects afterwards.
+      if (ptrType.getNumElements() == 1 &&
```
hongziqi, line 766: The current problem is irrelevant to this logic. It is recommended that a separate PR be submitted and the corresponding test cases be added.

elstainniles: test case:third_party/ascend/unittest/pytest_ut/test_loop_carried_mask.py
@pytest.mark.parametrize("compile_mode", ["simd", "simd_simt_template"])
def test_loop_singleton_mask(compile_mode):
    check_loop_mask([0, 2, 0], 1, observe=True, suffix=True, compile_mode=compile_mode, block=1)

WuTYSFG (approved):

hongziqi (approved):

PR #2076, opened 2026-09-09
[costmodel](fix) Scan op in loopCarriedRecurrence applies factor

AmagiriQwQ:
[sim_costmodel](fix) apply prefix-scan dependency factor only to the tt.scan portion of recurrence shuffle

## Why

`chunk_global_cumsum_scalar_kernel` (fla) was mis-routed to `all_simd` under costmodel auto even though SIMT is measurably faster on hardware. Its loop body carries a loop-carried dependency (`b_z`) together with `tt.scan` + `tt.reduce`, so the stage classifies as LoopCarriedRecurrence — the loop-carried dependency wins over hasPrefixScan, making the PrefixScan model unreachable. That model billed shuffle at the ideal vector rate and never consumed the measured `prefix_scan_dependency_factor`; only the unreachable PrefixScan model did.

The first fix (f37e1b217) multiplied the factor into the *whole* shuffle pool of the recurrence stage. That over-corrects: in this kernel `tt.scan` and `tt.reduce` each contribute exactly 50% of the shuffle steps, and a tree reduction does not suffer the scan's serialized lane-dependency chain (active lanes halve every level, so the N·log2(N) lane-step billing already carries enough slack). Shapes that should have stayed SIMD flipped to all_simt.

## What

- Split `StageWorkload.shuffleLaneSteps` by reduction source: `scanShuffleLaneSteps` records the `tt.scan` contribution, the remainder is `tt.reduce`. Both channels stay in sync across stage merge and loop-multiplicity scaling.
- LoopCarriedRecurrence now scales only the scan-class shuffle with the mode-specific `prefixScanDependencyFactor` (SIMD 17.35 / SIMT identity), semantically aligned with the standalone PrefixScan model.
- Route dump exposes `scan_shuffle_lane_steps_per_iteration` and `scan_shuffle_per_iteration` for calibration auditing.

## Effect (chunk_global_cumsum_scalar_kernel, stage_4 SIMD cycles, 4096-lane shape)

- before any fix: 84.6 — mis-routed to all_simd
- whole-pool factor: 1131.0 — over-corrected
- this change: 607.8 = 84.6 + 16.35 × 2048/64 — exact match with the
  scan-only prediction; 768-lane flips back to all_simd, 1792/4096-lane
  stay all_simt_only.

## Tests

- **C++ unit test — `SimdSimtCostModelTest.LoopCarriedRecurrenceAppliesScanDependencyFactor`** (new): a recurrence stage with a mixed shuffle pool (320 `tt.scan` + 320 `tt.reduce` lane-steps) shows that only the scan-contributed half consumes the dependency factor (+60 cycles over 4 iterations), that the SIMT side stays at the identity factor, and a reduce-only negative case proves the factor is not consumed when only `hasPrefixScan` is set.  `SimdSimtCostModel`: 35/35 pass.

- **End-to-end route guard — `third_party/ascend/unittest/costmodel_ut/test_costmodel_recurrence_scan_route.py`** (new): a self-contained kernel with a loop-carried recurrence and a nested `tl.cumsum` (structure mirrors fla `chunk_global_cumsum_scalar_kernel`) is compiled in auto mode with `auto_simt_scope_dump`; the test asserts both the numerics (against `torch.cumsum`) and that the dumped decision is a SIMT route (`all_simt_only` / `mixed_simd_simt`).  Sensitivity check: neutralising the SIMD `prefix_scan.dependency_factor` to 1.0 (equivalent to the pre-fix behaviour) makes the same kernel route to `all_simd`, i.e. the guard fails without this change.  Verified on 910_95; passes standalone (`python …`) and under `pytest`.

- **Existing route guards — `pytest_ut/test_simd_simt_costmodel_cases.py`**: 10/10 pass (needs `TRITON_RUN_SIMD_SIMT_COSTMODEL_GUARDS=1`), confirming the previously-calibrated routes are unchanged.

- **End-to-end**: fla `chunk_global_cumsum` auto route now selects `all_simt_only` (620.8 SIMD vs 428.4 SIMT system cycles).

# New contributor declaration
  `https://cbea.ms/git-commit/#why-not-how` .

  - [x] I have added tests.
    - `/unittest` for C++ tests (`SimdSimtCostModelTest.cpp`)

  - [ ] The `lit` tests I have added follow these `https://mlir.llvm.org/getting_started/TestingGuide/#filecheck-best-practices` , including the "tests should be minimal" section.

### DIFF
--- a/third_party/ascend/costmodel/include/AscendModel/RouteModel/StageRouteCostModel.h
+++ b/third_party/ascend/costmodel/include/AscendModel/RouteModel/StageRouteCostModel.h
@@ -90,6 +90,10 @@ struct StageWorkload {
   double storeWarpInstructions = 0.0;
   double predicateElements = 0.0;
   double shuffleLaneSteps = 0.0;
+  /// Portion of shuffleLaneSteps contributed by tt.scan (prefix-scan class).
+  /// The remainder is contributed by tt.reduce.  Only the scan portion
+  /// consumes the prefix-scan dependency factor inside recurrence stages.
+  double scanShuffleLaneSteps = 0.0;
   double dotFlops = 0.0;
   double issueElements = 0.0;
   double estimatedSpillTransactions = 0.0;
@@ -110,6 +114,8 @@ struct StageResourceCycles {
   double compute = 0.0;
   double predicate = 0.0;
   double shuffle = 0.0;
+  /// Cycles for the tt.scan-contributed portion of shuffle at the ideal rate.
+  double scanShuffle = 0.0;
   double dot = 0.0;
   double loopControl = 0.0;
   double branchControl = 0.0;
--- a/third_party/ascend/costmodel/lib/AscendModel/Analysis/StagePartitioner.cpp
+++ b/third_party/ascend/costmodel/lib/AscendModel/Analysis/StagePartitioner.cpp
@@ -165,7 +165,7 @@ static void accumulateDotWorkload(Operation *operation, StageWorkload &work) {
 }
 
 static void accumulateReductionWorkload(Operation *operation,
-                                        StageWorkload &work) {
+                                        StageWorkload &work, bool isScan) {
   if (operation->getNumOperands() == 0)
     return;
   auto input = dyn_cast<ShapedType>(operation->getOperand(0).getType());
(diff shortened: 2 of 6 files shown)


kaixin1976 (approved):

WuTYSFG (approved):

WuTYSFG:
请消除描述中的手动换行

PR #2078, opened 2026-09-09
[TritonToUnstructure](fix) prevent stale offset-cache hits

lichangdongtw:
## Problem / 问题

Compiling `fused_recurrent_gated_delta_rule` can hit the `combineInfo()` rank assertion while `ReplaceArguments` rebuilds loops carrying scalar/tensor pointers into base+offset form. Old IR is destroyed while `offsetMap` still references its Values; later allocations can reuse those addresses and incorrectly hit stale analysis.

编译 `fused_recurrent_gated_delta_rule` 时，`ReplaceArguments` 会重建携带标量/张量指针的循环，将指针改写为 base+offset。旧 IR 已被销毁，但 `offsetMap` 仍保留其 Value 的分析结果；后续新 Value 复用旧地址时，可能误命中过期记录并触发 `combineInfo()` 的 rank 断言。

## Fix / 修复

After redirecting result uses, move replaced operations into a detached local `Block retiredOps`. The old operations leave the function immediately, while their Values remain allocated until rewriting finishes. Clear `offsetMap` before the block's destructor destroys the retired IR. This prevents address reuse during analysis and preserves cache reuse across replacements. Retaining old IR and cached analysis can increase peak memory.

替换结果的使用后，将旧操作移入脱离函数的局部 `Block retiredOps`。旧操作立即从函数中移出，但其 Value 的存储保留到改写结束；先清空 `offsetMap`，再由该块的析构函数销毁旧 IR。这样既阻止分析期间的旧地址复用，也保留各次替换之间可复用的缓存。保留旧 IR 和分析缓存可能增加峰值内存。

### Why this approach / 为什么采用这个方案

We compared three other approaches:

此前比较过以下三种方案：

- **Full cache clearing:** clearing `offsetMap` before each deletion fixes the lifetime issue, but also discards unaffected analysis, including reusable calculations outside the rewritten loop.

  **整表清理：**每次删除旧 IR 前清空 `offsetMap` 可以修复生命周期问题，但也会丢弃不受影响的分析结果，包括循环外原本可以复用的计算。

- **Selective invalidation:** filtering must account for cached payloads and transitive SSA dependencies. The first version missed dependencies on earlier SCF results; the strict version required recursive dependency checks and a cache scan at each rewrite. In our shared-prefix synthetic case, that checking cost outweighed the saved analysis.

  **选择性清理：**筛选需要覆盖缓存载荷和传递性的 SSA 依赖。初版遗漏了对前序 SCF 结果的依赖；严格版需要递归检查依赖，并在每次改写时扫描缓存。在共享前置计算的人工用例中，筛查成本超过了省下的重复分析成本。

- **Epoch validation:** checking entry generations allows lazy invalidation, but requires a new cache container, stability classification, and a versioning protocol. The tested implementation changed five production files.

  **Epoch 校验：**通过检查缓存记录的版本来实现延迟失效，但需要新增缓存容器、稳定性分类和版本管理协议；已测试的实现涉及五个生产文件。

Deferred destruction addresses address reuse directly through object lifetimes, keeps the existing cache representation, and limits the production change to one implementation file.

延迟销毁直接从对象生命周期上阻止地址复用，沿用现有缓存结构，生产代码只需修改一个实现文件。

### Why not leave cleanup to CSE / 为什么不把旧 IR 留给 CSE 清理

Redirecting result uses does not make a loop with side effects dead. In a small MLIR check, both CSE and CSE followed by Canonicalizer removed an unused pure loop, but retained a loop containing a store. When two copies of the store-carrying loop were present, both loops and both stores remained. Leaving old and replacement loops in the function would therefore retain duplicate memory writes.

替换结果的使用，并不会让带副作用的循环变成可删除的死代码。在小型 MLIR 实验中，CSE 以及 CSE 后接 Canonicalizer 都能删除未使用的纯计算循环，但会保留含有 store 的循环；同时放入两份这样的循环时，两份循环和两个 store 都被保留。因此，把旧循环和新循环都留在函数中，会留下重复的内存写操作。

The old operations must leave the function without being freed while the cache can still reference them. The detached `retiredOps` block owns them during this interval: they are absent from the rewritten function and its traversal, and are destroyed only after the final cache clear. Cleanup therefore has an explicit lifetime boundary rather than depending on a later optimization pass.

所以，需要在缓存仍可能引用旧 Value 时，将旧操作移出函数，同时暂缓释放。脱离函数的 `retiredOps` 在这段时间内持有旧操作：它们不再出现在改写后的函数及其遍历中，直到最后清空缓存后才被销毁。这样清理时机由明确的生命周期边界保证，无需依赖后续优化 pass。

## Tests / 测试

- **C++ test: cache lifetime.** A 13-line IR input builds a loop carrying a pointer argument. The test calls the real analysis and rewrite functions and checks that the cache no longer references the old Value when its storage is released. It reliably detects the original bug and an early-deletion variant without relying on chance address reuse.

  **C++ test：验证缓存生命周期。** 用 13 行 IR 构造一个携带指针参数的循环，调用真实分析和改写函数，在旧 Value 的存储释放时检查缓存已不再引用它。它不依赖随机地址复用，能稳定检出原始错误实现和提前删除变体。

- **MLIR test: rewrite result.** A 20-line IR input builds two nested loops carrying both scalar and tensor pointers. The test runs the actual conversion pass, then uses FileCheck to verify offset propagation and that only the expected loops and stores remain, checking for leftover old operations.

  **MLIR test：验证改写结果。** 用 20 行 IR 构造同时携带标量、张量指针的两层嵌套循环，运行实际转换 pass，再通过 FileCheck 检查偏移传递是否正确，以及输出中是否只有预期的循环和 store，避免旧操作残留。

Both tests are integrated into the existing Ascend CI, which runs CTest before lit.

两项测试均接入现有 Ascend CI，先运行 CTest，再运行 lit。

## New contributor declaration

- [x] This is a nontrivial fix.
- [x] The description explains the cause and rationale.
- [x] Run `pre-commit run --from-ref fcc41c37a8302f69378b834376843042f50a3668 --to-ref HEAD` on the final submission commit.
- [x] Added C++ and focused MLIR regression tests.
- [x] The added hand-written lit input focuses on the relevant behavior.

### DIFF
--- a/third_party/ascend/lib/TritonToUnstructure/ReplaceArguments.cpp
+++ b/third_party/ascend/lib/TritonToUnstructure/ReplaceArguments.cpp
@@ -708,6 +708,7 @@ void replacePtrArguments(triton::FuncOp funcOp,
         rewriter.replaceAllUsesWith(res, replacement);
         std::advance(resIter, std::max(getPtrTensorRank(res.getType()), 1));
       }
+      offsetMap.clear();
       rewriter.eraseOp(op);
       op = newOp;
       convertTensorPtrPre(op, rewriter, offsetMap);


third_party/ascend/lib/TritonToUnstructure/ReplaceArguments.cpp
```diff
@@ -708,6 +708,7 @@ void replacePtrArguments(triton::FuncOp funcOp,
         rewriter.replaceAllUsesWith(res, replacement);
         std::advance(resIter, std::max(getPtrTensorRank(res.getType()), 1));
       }
+      offsetMap.clear();
```
WuTYSFG, line 711: 能提供可复现的最小用例吗

TecJesh, line 711: 代码变更需要往main-dev提pr，目前只有文档变更可以直接合入main

lichangdongtw: 感谢提醒，已将原 PR 改投 `main-dev`，并基于 `fcc41c37a` 重新应用、验证了修复。最终仍只有 `ReplaceArguments.cpp` 增加一行 `offsetMap.clear();`，PR 描述已更新为这一轮的验证结果。

lichangdongtw: 已补充[独立编译器复现用例](https://gist.github.com/lichangdongtw/3719038b2f140c701af1b8fdcff067ad)（[下载 ZIP](https://gist.github.com/lichangdongtw/3719038b2f140c701af1b8fdcff067ad/archive/842eb4d04285301937c6873af968b34fc0f75e73.zip)），包含 193 行 TTIR、两个 Python 脚本，以及环境说明、错误日志和逐轮结果。解压进入目录后运行：

```bash
python run_trials.py --runs 100 --output baseline-results
```

已在 `main-dev` 的 `fcc41c37a` 上重新构建对照：这轮修复前 100 次出现 3 次相同 rank 断言，修复后 100 次全部通过且输出 IR 一致。

KanuaK, line 711: Add test case for fixed situation to pytest_ut (or MLIR test case to unittest/Conversion), which ensure that your fix can resolve the particular problem and future PRs won't make your fix invalid.

lichangdongtw: 已补充小型仓库内回归测试：包含 13 行 IR 的 C++ 生命周期测试，以及 20 行输入的嵌套循环 MLIR 用例。此前 Gist 保留为原始问题的独立复现材料。

Added small in-tree regressions: a C++ lifetime test with 13 lines of IR and a nested-loop MLIR test with a 20-line input. The earlier Gist remains the standalone reproducer for the original failure.

lichangdongtw: 已新增 C++ 生命周期测试，在旧 Value 的存储释放时检查缓存不再引用它；原始错误实现与提前删除变体都会被检出，不依赖地址复用。另加普通 MLIR 回归，检查嵌套循环中的标量/张量指针改写，以及输出 IR 中没有残留重复的旧循环或 store。两项测试已接入现有 CI。

Added a C++ lifetime regression that checks that the cache no longer references an old Value when its storage is released; both the original bug and an early-deletion variant fail without relying on address reuse. A focused MLIR regression covers scalar/tensor pointer rewriting in nested loops and checks that no duplicate old loops or stores remain in the output IR. Both tests are integrated into the existing CI.

third_party/ascend/unittest/offset_analysis_ut/OffsetCacheLifetimeTest.cpp
```diff
@@ -0,0 +1,119 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+#include "TritonToUnstructure/UnstructureConversionPass.h"
```
hongziqi, line 23: rename dir to `Analysis`

WuTYSFG (approved):

hongziqi (approved):

hongziqi (approved):

PR #2087, opened 2026-09-09
[ssbuffer](fix) do not erase live ops or stale manager entries

CrDem:
eraseOpsWithUnusedUsers removed the yield's defining op unconditionally, on the assumption that rewiring the yield onto the clone had dropped its last use. A loop-carried value is regularly consumed inside the body as well, so the assumption does not hold and the remaining consumers were left with dangling operands. It also counted uses on getResult(0) rather than on the result that was actually the operand, which reports a multi-result op as dead while other results are live.

ComputeBlockIdManager caches raw Operation pointers in blockIdToOps and opToBlockId and had no way to invalidate them, so erased operations stayed in the maps. cloneDepSubgraph and willCreateCycle dereference that vector through getOpsByBlockId, and the manager is constructed once per module, so an erase performed for one main loop corrupts the next. Add forgetOp and withdraw an operation before erasing it.

# New contributor declaration

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests
  - [x] This PR does not need a test because `the defect is a use-after-free, not a difference in emitted IR. With the bug present the pass dereferences freed memory, which surfaces as a crash whose symptom depends on the allocator rather than as a stable diagnostic; with it fixed, the IR produced for any input that previously survived is byte-identical. There is therefore nothing for FileCheck to assert on. Built and exercised on a flash-attention kernel that goes through the dynamic CV pipeline; generated code is unchanged for inputs that previously compiled, and a case that previously aborted inside RefineArgsBlockIdPass now completes`.

### DIFF
--- a/third_party/ascend/include/DynamicCVPipeline/PlanComputeBlock/ComputeBlockIdManager.h
+++ b/third_party/ascend/include/DynamicCVPipeline/PlanComputeBlock/ComputeBlockIdManager.h
@@ -52,6 +52,8 @@ public:
   void updateBlockIdWithInner(Operation *parentOp, int targetId);
   void updateBlockId(Operation *op, int blockId);
 
+  void forgetOp(Operation *op);
+
   bool shouldInheritFromParent(Block *block, CoreType requiredCoreType) const;
   llvm::LogicalResult inheritFromParent(Block *block);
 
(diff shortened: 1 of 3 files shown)


VDV1985 (approved):
/LGTM

WuTYSFG (approved):

Skyminers (approved):

hongziqi (approved):

PR #1988, opened 2026-09-05
[simt_costmodel](fix) pass module to set_buffer_count

swordfate:
# New contributor declaration
- [x] I am not making a trivial change, such as fixing a typo in a comment.

- [x] I have written a PR description following these
  [rules](https://cbea.ms/git-commit/#why-not-how).

- [x] I have run `pre-commit run --from-ref origin/main --to-ref HEAD`.

- Select one of the following.
  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests
    - `/python/test` for end-to-end tests
  - [x] This PR does not need a test because it restores the required `mod` argument in
         existing `set_buffer_count` calls.

- Select one of the following.
  - [x] I have not added any `lit` tests.
  - [ ] The `lit` tests I have added follow these [best practices](https://mlir.llvm.org/getting_started/TestingGuide/#filecheck-best-practices),
    including the "tests should be minimal" section. (Usually running Python code
    and using the instructions it generates is not minimal.)

## Why

`ascend.passes.ttir.set_buffer_count` expects the MLIR module as the first
argument. The current calls omit `mod`, so configuring `intra_cache_num`,
`inter_cache_num`, or `load_cache_num` can fail in the backend.

## What

Pass `mod` to all three `set_buffer_count` calls:

```python
ascend.passes.ttir.set_buffer_count(mod, "INTRA", _intra_val)
ascend.passes.ttir.set_buffer_count(mod, "INTER", _inter_val)
ascend.passes.ttir.set_buffer_count(mod, "LOAD", _load_val)

PR #2101, opened 2026-09-10
[simt_costmodel](fix) re-enable AutoBlockify for all-SIMD routes

swordfate:
# New contributor declaration

  - [ ] I have added tests.
    - [ ] `/test` for `lit` tests
    - [ ] `/unittest` for C++ tests
    - [ ] `/python/test` for end-to-end tests
  - [x] This PR does not need a test because it restores the pre-6e29d8e
         AutoBlockify V1 flag policy while keeping the existing linalg-level
         blacklist contract; the existing costmodel compiler contract tests
         are unaffected and stay green.

## Why

The 6e29d8e refactor made `--enable-auto-blockify-loop` conditional on
`parallel_mode == mix_simd_simt` and also force-disabled V1 for `all_simd`.
This prevents all-SIMD and ordinary SIMD kernels from using AutoBlockify even
when the shared policy allows it, while the current cost model scores
`all_simd` from a post-AutoBlockify analysis view.

## What

- Keep the resolved V1 policy for `all_simd` instead of force-disabling it.
- Stop gating the NPUIR `--enable-auto-blockify-loop` flag on `parallel_mode`,
  restoring the pre-6e29d8e behavior for every route.
- Honor the linalg-level blacklist (`has_auto_blockify_blacklist_op`) in both
  bin paths.  `DiscreteMaskAccessConversion` inserts `hivm.sync_block_lock`
  when it cannot prove program-id non-overlap (e.g. discrete masked stores
  over dynamic indices), and the launcher then deliberately disables the
  launch-grid cap:
  `enable_auto_map_parallel_blocks = auto_blockify_v1_runtime_cap && !has_auto_blockify_blacklist_op`.
  Requesting the NPUIR persistent loop without that cap deadlocks, so the
  compiler must exclude these kernels exactly like the legacy path does
  (`_is_auto_map_parallel_blocks_enabled() && !has_auto_blockify_blacklist_op`).
- Only mixed/local-scope routes carry `--super-block-factor`; `all_simd`
  remains F1, matching the historical SIMD behavior.

Only `third_party/ascend/backend/compiler.py` is touched (+36/-27).

## Tests

No new UT is included in this PR.  On the branch, the existing costmodel
contract test was run and stays green:

```text
python third_party/ascend/unittest/costmodel_ut/test_compiler_costmodel_contract.py \
  CompilerCostmodelContractTest.test_all_simd_decision_replaces_route_request_with_backend_mode
-> OK
```

On an Ascend A3 test machine, for a kernel that hits the linalg blacklist
(`test_triton_unified_attn[...]` with dynamic-index masked stores):

- base (no flag requested): passes;
- this change without the blacklist check (previous revision): fails with NPU
  aicore timeout, because the pass is requested while the launcher disables
  the grid cap;
- this revision: the blacklisted kernel does not request the pass, while
  non-blacklisted kernels still get the global V1 behavior.

CI integration tests are re-run by this push.

### DIFF
--- a/third_party/ascend/backend/compiler.py
+++ b/third_party/ascend/backend/compiler.py
@@ -187,12 +187,11 @@ def _apply_cpp_simd_simt_decision(metadata, effective: str, superblock_factor: i
         # stages add mixed-runtime flags even though no SIMT scope exists.
         metadata["compile_mode"] = "simd"
         metadata["parallel_mode"] = "simd"
-        # SIMT AutoBlockify V1 uses the SIMT block/thread execution contract.
-        # It may be present on the analysis copy so StagePartitioner can see
-        # its scheduling loop, but an all-SIMD executable must launch the
-        # original logical grid and must not request the NPUIR V1 pass.
-        metadata["auto_blockify_v1_enabled"] = False
-        metadata["auto_blockify_v1_runtime_cap"] = False
+        # All-SIMD can still execute AutoBlockify V1 as a logical-program loop.
+        # Keep the policy resolved before the cost model instead of disabling it.
+        v1_materializable = bool(metadata.get("route_transform_v1_materializable", False))
+        metadata["auto_blockify_v1_enabled"] = v1_materializable
+        metadata["auto_blockify_v1_runtime_cap"] = v1_materializable
     elif effective == "mixed_simd_simt":
         metadata["auto_simt_requested_kind"] = effective
         # Mixed execution still needs V1 at factor one: its persistent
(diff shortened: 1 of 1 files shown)


KanuaK (approved):

hongziqi (approved):

third_party/ascend/backend/compiler.py
```diff
@@ -1507,15 +1512,19 @@ def linalg_to_bin_enable_npu_compile_A2_A3(linalg: str, metadata, opt):
...
-        if npuir_v1_mixed_enabled:
-            selected_factor = _selected_npuir_superblock_factor(metadata, opt)
-            _compile_option_list += [
-                "--enable-auto-blockify-loop",
-                f"--super-block-factor={selected_factor}",
-            ]
+        # Honor the linalg-level blacklist (e.g. hivm.sync_block_lock) as well:
+        # the launcher disables the grid cap for blacklisted kernels, so
+        # requesting the NPUIR persistent loop without that cap would deadlock.
+        npuir_v1_enabled = (metadata.get("auto_blockify_v1_enabled", False)
```
KanuaK, line 1518: modify variable name with auto_blockify

swordfate: I modified the name `npuir_v1_enabled` to `auto_blockify_npuir_v1_enabled ` in https://github.com/triton-lang/triton-ascend/pull/1988/

PR #2130, opened 2026-09-12
[backend](revert) restore AutoBlockify compiler and launcher gates

zhanwei33:
Restore the original AutoBlockify compiler-option conditions and launcher block-count cap. In the supplied cumsum regression, the good and bad kernels have byte-identical TTIR, TTAdapter, BCMLIR, and npubin, but the bad launcher omits the original `min(blockNum, 56)` cap. RowCoalescing makes centralized `auto_blockify_enabled=False`, so restoring the compiler flags alone does not restore the launch behavior.

- Restore the three `--enable-auto-blockify-loop` conditions at their original option-assembly sites: backend policy plus blacklist for Linalg, with the existing RowCoalescing gate for pure SIMT.
- Restore the launcher policy/blacklist decision at its original site and use it for block-count capping, including after RowCoalescing as in the original revision.
- Preserve `_finalize_program_launch_policy(metadata, opt)` and existing metadata production, checks, and defaults, including `ptsm_cap_authorized`.
- Update the three regression-test expectations to match the restored policy. Keep all 32 compiler argv combinations, check RowCoalescing ceil-div and capping, and cover all four backend-policy/blacklist combinations with either centralized metadata value in both launch APIs.

Validation: the two source-loaded Python suites pass on this branch (**20 passed, 28 existing skips**), and pre-commit and `git diff --check` pass. The latest commit changes only the two test files. Prior artifact validation showed that restoring the launcher adds exactly the block-count cap in both launch APIs for the supplied bad metadata; the historical compiler argv and launcher policy matrices also passed.

Local validation uses source-loaded modules and mocked compiler/backend interactions. No local wheel build or NPU correctness/performance rerun was performed; the supplied artifacts do not quantify recovered performance.

Base: `main` at `3841a37338b7e342f2a88d95818c81274c947040`. Supersedes #2127.

### DIFF
--- a/third_party/ascend/backend/compiler.py
+++ b/third_party/ascend/backend/compiler.py
@@ -842,7 +842,7 @@ def linalg_to_bin_enable_npu_compile_910_95(linalg: str, metadata, opt):
                 _compile_option_list += \
                     [f"--link-aicore-bitcode={bitcode}"]
 
-        if metadata["auto_blockify_enabled"]:
+        if _is_auto_map_parallel_blocks_enabled() and not metadata.get("has_auto_blockify_blacklist_op", False):
             _compile_option_list += ["--enable-auto-blockify-loop"]
         npu_compiler_path, env = _get_npucompiler_path()
         if npu_compiler_path.endswith("bishengir-compile"):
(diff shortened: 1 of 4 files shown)


LH-123L (commented):
oK

insanecoder-zzj (approved):

KanuaK (approved):

WuTYSFG (approved):

PR #2135, opened 2026-09-12
[costmodel](fix) extend SuperBlock factors through 32

Huangdinosaur:
# feat: extend SuperBlock factors through 32

SuperBlock factor support was limited to 1, 2, and 4 across several independent checks, which prevented the cost model and the all-SIMT AutoBlockify path from considering larger legal groups. This change extends the modeled factor range through 32 while retaining the hardware constraint that `numWarps * factor <= 64`.

- centralize supported factors as 1, 2, 4, 8, 16, and 32
- filter candidates with the `numWarps * factor <= 64` constraint
- extend all-SIMT and mixed route modeling and TA materialization
- propagate the selected all-SIMT factor to `bishengir-compile`
- keep mixed backend execution capability capped at factor 4
- add policy and IR coverage, and validate cost-model and NPU runtime behavior

The mixed route remains capped at factor 4 when publishing executable backend capability because the current NPUIR local-scope materializer accepts only factors 1, 2, and 4. The upper-level route model and TA materialization interfaces are ready for larger mixed factors once that lower-level restriction is removed and validated.

Validation:

- 29 AutoBlockify policy tests passed
- F32 direct materialization and F1-to-F32 refinement IR checks passed
- invalid F3 and F64 IR checks passed
- the legal `numWarps`/factor candidate matrix passed
- 54 legal all-SIMT `numWarps`/factor/materializer NPU runtime combinations passed
- 10 existing SIMD/SIMT cost-model end-to-end tests passed, including an all-SIMT case that selected factor 8

# New contributor declaration

  - [x] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests


Huangdinosaur:
retry

kaixin1976 (approved):

hongziqi (approved):

PR #2161, opened 2026-09-14
[ssbuffer](fix) catch inter-block memref deps in CUBE scope

CrDem:
AddMultiBufferInnerScope rejects memref-typed values crossing a block boundary only in VECTOR scopes. A CUBE main loop still goes through CloneOps and the stage split, where such a dependency keeps a cloned sync op alive and duplicates the handshake across pipeline stages: wrong results or a hang at an inter-core buffer count of 2. Apply the same check to CUBE scopes and fall back.

validateClonedSyncOpsErased only tested top-level op types, so a surviving clone wrapped in the multi-buffering scf.if went unnoticed. Look inside cloned ops so that case falls back too.

# New contributor declaration

  - [x] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests


VDV1985 (approved):
/LGTM

Skyminers (approved):

insanecoder-zzj (approved):

PR #2165, opened 2026-09-15
[frontend](fix) drop read/write effects mark on buffer-language alloc

jqran:
### What this fixes

`bl.alloc()` unconditionally attached an `effects = ["write", "read"]` mark to every buffer it allocated, so each alloc got a synthetic read user even when it is never written.

PlanMemory in AscendNPU-IR derives a buffer's lifetime (gen at first write, kill at last read) from the real memory effects in the IR, so the bogus read makes it treat an uninitialized alloc as a live buffer: the buffer survives the first write, and planning then reports a use of a buffer that has not been generated yet. Only the caller knows whether a buffer is written, so annotate nothing but `mem_unique` here and let the passes derive the effects of each buffer from the actual stores and loads.

### How it was verified

- Built on x86 (ubuntu20 container) from `main-dev` + this patch: wheel builds clean.
- IR: `memref.alloc` now carries only `annotation.mark {mem_unique}`; the `effects` attribute is gone. `is_mem_unique=True` is still annotated.
- New `third_party/ascend/unittest/pytest_ut/test_alloc_annotation.py` (3 cases, all passing on a5):
  `test_alloc_does_not_annotate_memory_effects` checks the TTIR (`mem_unique` kept, `effects` gone) and fails when
  `semantic.py` is reverted to the previous behaviour, so it is a real regression guard;
  `test_alloc_as_fixpipe_dst_on_npu` / `test_unique_alloc_as_fixpipe_dst_on_npu` really run a `bl.alloc`ated UB
  buffer the way the flash attention kernel uses it (fixpipe destination of a `tl.dot` result, read back with
  `to_tensor`) and compare against torch on the NPU.
- a5 (UB/NPU 1), this branch: `test_alloc_annotation.py` 3 passed; with the rest of the buffer-language set
  (`test_alloc`, `test_buffer_language`, `test_fixpipe`, `test_fixpipe_tensor_dst`, `test_multibuffer`,
  `test_scope_buffer`, `test_to_tensor`, `test_bind_buffer`, `test_copy`) 19 passed.
- x86 ubuntu20 container (no torch, so the two NPU cases are skipped there): 1 passed, 2 skipped.

# New contributor declaration

  - [x] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests
      -> `third_party/ascend/unittest/pytest_ut/test_alloc_annotation.py` (runs the
         kernels on the NPU and checks the generated TTIR)


python/triton/extension/buffer/language/semantic.py
```diff
@@ -44,7 +44,6 @@ def alloc(etype: tl.dtype, shape: List[tl.constexpr], address_space: bl.address_
     handle = builder.alloc(memref_ty)
     if is_mem_unique:
         builder.create_annotation_mark(handle, "mem_unique", builder.get_unit_attr())
-    builder.create_annotation_mark(handle, "effects", builder.get_str_array_attr(["write", "read"]))
```
hongziqi, line 47: add test case

jqran: Added in 31f4a6df3: `third_party/ascend/unittest/pytest_ut/test_alloc_annotation.py`

- `test_buffer_language_alloc_on_npu` really runs two `bl.alloc` kernels on the NPU (plain and `is_mem_unique=True`) and compares the result against torch, so this change cannot break code generation silently.
- `test_alloc_does_not_annotate_memory_effects` checks the TTIR of the same kernels: no `annotation.mark` may carry an `effects` attribute, while `is_mem_unique=True` must still be annotated with `mem_unique` (that assertion fails on the previous behaviour).

Verified locally on the same commit: a5 (UB/NPU 1) 19 passed for the buffer-language set (`test_alloc_annotation`, `test_alloc`, `test_buffer_language`, `test_fixpipe`, `test_fixpipe_tensor_dst`, `test_multibuffer`, `test_scope_buffer`, `test_to_tensor`, `test_bind_buffer`, `test_copy`), x86 ubuntu20 container 8 passed. Reverting `semantic.py` to the old behaviour makes the annotation test fail, so it is a real regression guard.

jqran: Updated in 3618acb6f — `third_party/ascend/unittest/pytest_ut/test_alloc_annotation.py` now has:

- `test_alloc_does_not_annotate_memory_effects`: checks the TTIR of kernels calling `bl.alloc` — no `annotation.mark` may carry an `effects` attribute, and `is_mem_unique=True` must still be annotated with `mem_unique`. The `mem_unique` assertion fails when `semantic.py` is reverted to the previous behaviour, so it guards exactly this change.
- `test_alloc_as_fixpipe_dst_on_npu` / `test_unique_alloc_as_fixpipe_dst_on_npu`: a `bl.alloc`ated UB buffer is used the way the flash attention kernel uses it (fixpipe destination of a `tl.dot` result, read back with `to_tensor`) and the result is compared against torch on the NPU.

Local results: `1 passed, 2 skipped` on a5 — the two end-to-end cases skip because the bishengir installed on my test machine cannot compile that fixpipe attribute layout (`expected mlir::hivm::FixpipeDualDstModeAttr, but got: #hivm.dma_mode<nz2nd>`), which is unrelated to this change. On a toolchain where that fixpipe lowering works, the same kernels ran and matched torch (max_abs_diff 7.6e-06 for the plain alloc).

jqran: Correction on my previous comment: the two end-to-end cases were being skipped because of a **test-environment mistake**, not a toolchain limitation.

`_get_npucompiler_path()` prefers `<site-packages>/triton/backends/ascend/bishengir/bin/bishengir-compile`, and my isolated install (pip `--target`) did not carry that directory, so compilation silently fell back to the 1.1.0 bishengir-compile on `PATH`, which rejects the fixpipe attribute layout. After pointing the test environment at the packaged bishengir-compile 1.2.0, the same kernels compile and match torch:

```
$ python -m pytest -q --noconftest -rs third_party/ascend/unittest/pytest_ut/test_alloc_annotation.py
3 passed
```

The skip branch has been removed, so all three cases in `3618acb6f`→`c3ddb0144` are hard assertions now.

hongziqi (approved):

KanuaK (approved):

PR #2177, opened 2026-09-15
[compiler](fix) preserve deprecated options for compatibility

KanuaK:
Keep legacy option names and defaults in NPUOptions so Inductor and JIT recognize them as compile options instead of kernel arguments.

Normalize deprecated user values on a copy while preserving alias routing and internal option round trips. Consume the legacy stream keyword at the Ascend boundary.

Deprecate enable_vf_fusion and stop forwarding its compiler flag. Add option parsing and JIT/Inductor regression coverage.


WuTYSFG (dismissed):

hongziqi (dismissed):

third_party/ascend/unittest/pytest_ut/test_compiler.py
```diff
@@ -121,3 +138,124 @@ def test_get_simt_stack_limit_config_overrides_default(monkeypatch, tmp_path):
...
+    raw = {"enable_mixed_cv": True, "enable_dynamic_cv_pipeline": False, "num_warps": 16}
+
+    options = backend.parse_options(raw)
+
+    assert options.enable_mixed_cv is True
+    assert options.enable_dynamic_cv_pipeline is False
+    assert options.num_warps == 16
+
+
+_IGNORED_OPTIONS = sorted(utils._DEPRECATED_NPU_OPTIONS - utils._DEPRECATED_NPU_OPTION_ROUTES.keys() -
```
heroisme6, line 163: Issue: In Python 3, `dict.keys()` returns a `dict_keys` view. While the operation `set - dict_keys` is valid, chaining subtraction operations (e.g., `a - b - c`) reduces readability; furthermore, it is prone to raising a `TypeError` if the collection is re-wrapped as a non-set type in a different module.

Recommendation: It is recommended to explicitly convert the view to a `set` and combine collections—using parentheses or a single-line expression—before performing the set difference:

_EXCLUDED_KEYS = (
    set(utils._DEPRECATED_NPU_OPTION_ROUTES)
    | set(utils._DEPRECATED_NPU_OPTION_ALIASES)
    | {"stream"}
)
_IGNORED_OPTIONS = sorted(set(utils._DEPRECATED_NPU_OPTIONS) - _EXCLUDED_KEYS)

third_party/ascend/unittest/pytest_ut/test_compiler.py
```diff
@@ -121,3 +138,124 @@ def test_get_simt_stack_limit_config_overrides_default(monkeypatch, tmp_path):
     result = compiler.get_simt_stack_limit()
     assert result == 9999
     assert result != 1152
+
+
+def test_empty_options_expose_all_legacy_names(backend):
+    options = backend.parse_options({})
+
+    assert utils._DEPRECATED_NPU_OPTIONS <= options.__dict__.keys()
+    assert options.__dict__["arch"] == options.arch == backend.target.arch
```
heroisme6, line 147: Issue: The code relies heavily on `options.__dict__` for iteration, value retrieval, and comparison. If `options` is subsequently modified to use `__slots__`, converted to a `dataclass(slots=True)`, or implemented using custom attribute proxies (Descriptors or Properties), `options.__dict__` will either fail to work or cease to reflect the complete set of attributes.
Recommendation: Prefer using `hasattr(options, key)` or a dedicated export method (such as `asdict()` or `to_dict()`) to check for the existence of fields. For value retrieval, prioritize direct attribute access (e.g., `options.arch`) to minimize reliance on direct manipulation of `__dict__`.

hongziqi (approved):

insanecoder-zzj (approved):

PR #2181, opened 2026-09-15
[ssbuffer](fix) always build for-loop polling cond at for body start

1801ZDL:
## Fix "operand #0 does not dominate this use" in AddMultiBufferOuterScope

**Problem**: for for-loops, the polling condition (`divsi/c2/remsi/c0/cmpi`) was emitted at the "anchor wait" position. The anchor election cannot compare waits across different blocks, so the anchor is not guaranteed to dominate every group's wrap — and when the anchor wait itself sits inside a user `scf.if`, the `(block, iterator)` insertion point is invalid outright. Both cases hit e2e (nsa_fwd / nsa_bwd) as `operand #0 does not dominate this use`.

**Fix**: always build the polling condition at the **start of the for body** (mirrors the while branch). Body start dominates every group's wrap at any nesting depth. The anchor wait now only provides the `tid` tag; `block_id` comes from the first body op (`getFirstBlockId`), same as the while branch.

**Verification**: nsa_fwd / nsa_bwd `inner`/`outer`/`inner+outer` all pass the verifier; determinism 10–20 runs single md5; AllocMultiCache LIT 8/8; e2e validated. For kernels with all body-level waits (sdf02-style) the only change is the condition chain moving to the top of the body — semantically equivalent, it only reads the induction variable and step.


third_party/ascend/lib/DynamicCVPipeline/AllocMultiCache/AddMultiBufferOuterScope.cpp
```diff
@@ -1319,9 +1319,15 @@ static Value prepareLoopPolling(Operation *loopOp, Operation *waitOp,
   int tid = getTransferId(waitOp);
 
   if (auto forOp = dyn_cast<scf::ForOp>(loopOp)) {
-    OpBuilder condBuilder(forOp.getBody(), Block::iterator(waitOp));
-    Value cond = createPollingCondition(forOp, condBuilder, bid, tid);
-    builderOut.setInsertionPoint(forOp.getBody()->getTerminator());
+    // Build the polling cond at body start so it dominates every group's
+    // scf.if wrapper (mirrors the while branch). tid comes from the anchor
+    // wait; the block id from the first body op (CloneOps contiguity).
+    builderOut.setInsertionPointToStart(forOp.getBody());
```
olivervnc, line 1325: 直接改 condBuilder 的 insertionPoint，如果不确定之后是否会用，就不要改外部的

third_party/ascend/lib/DynamicCVPipeline/AllocMultiCache/AddMultiBufferOuterScope.cpp
olivervnc, line 1359: 遗留：for和while的重复逻辑提取公共函数，计数器 -> condition

third_party/ascend/lib/DynamicCVPipeline/AllocMultiCache/AddMultiBufferOuterScope.cpp
olivervnc: 遗留：确认之后是否会不改 InsertionPoint 就使用 builderOut

olivervnc (approved):

hongziqi (approved):

KanuaK (approved):

Issue #2201, opened 2026-09-16
[Bug]: Ascend launcher rejects negative signed int8 scalar arguments


luochao-dev:
### Describe the bug

On `release/3.2.2`, the generated Ascend Python launcher rejects valid
negative scalar arguments annotated as `tl.int8` before the kernel is
launched.

Minimal reproducer:

```python
import torch
import triton
import triton.language as tl


@triton.jit(do_not_specialize=["value"])
def signed_int8_kernel(output, value: tl.int8):
    tl.store(output, value.to(tl.int32))


output = torch.empty(1, dtype=torch.int32, device="npu")
signed_int8_kernel[(1,)](output, -113)
```

Actual result:

```text
OverflowError: unsigned byte integer is less than minimum
```

Expected result:

The launcher should accept values in the signed `int8` range `[-128, 127]`,
and the kernel should store `-113`. Values outside that range should raise
`OverflowError` without truncation.

The failure happens in the generated CPython launcher before entering the NPU
launch path. `generate_npu_wrapper_src()` maps `int8_t` to the
`PyArg_ParseTuple` format unit `b`, but CPython defines `b` as an unsigned byte.
There is no signed-byte format unit. Changing only the format to `i` would not
be safe because `i` writes through an `int*`, while the generated destination
is an `int8_t`.

### Environment details

- Triton-Ascend branch: `release/3.2.2`
- Python: 3.10
- Backend: Ascend NPU
- The exception is raised by CPU-side argument parsing before any NPU runtime
  call, so it is independent of the NPU model and CANN execution path.

luochao-dev:
A fix for `release/3.2.2` is proposed in #2202.

hongziqi:
This issue also exists in upstream Triton 3.2 NVIDIA/AMD launchers because signed i8 is parsed using the unsigned PyArg_ParseTuple("b") format. Upstream Triton 3.7 resolves it as part of the launcher refactoring by parsing integer arguments directly with PyLong_AsLong(https://github.com/triton-lang/triton/pull/6788/changes).
﻿
To avoid introducing an Ascend-specific parsing path and behavior difference in release/3.2.2, we plan not to backport this fix independently. Instead, the issue will be resolved by synchronizing the upstream 3.7 launcher optimization into the Ascend 3.7 branch, with a signed-i8 regression test added there.

PR #2202, opened 2026-09-16
[launch](fix) parse signed int8 scalar arguments

luochao-dev:
## Summary

- Fix signed `int8` scalar parsing in the Ascend Python launcher.
- Use a `PyArg_ParseTuple` `O&` converter because the `b` format parses an
  unsigned byte and rejects valid negative `int8` values.
- Validate the `[-128, 127]` range before writing to `int8_t`.
- Add one launcher source-generation regression test.

## Background

On `release/3.2.2`, passing a negative scalar annotated as `tl.int8` fails
during CPU-side argument parsing with:

```text
OverflowError: unsigned byte integer is less than minimum
```

The failure occurs before entering the NPU launch path. `O&` allows the
launcher to parse through a custom converter without writing an `int` into an
`int8_t` destination.

Fixes #2201.

## Validation

- All 4 launcher export test functions passed through dependency-isolated
  execution.
- Python syntax checks passed.
- `git diff --check` passed.
- Standalone CPython C++ validation accepted `-128`, `-113`, `-1`, `0`, and
  `127`, and rejected `-129` and `128` with `OverflowError`.
- Full NPU runtime validation was not available in the current environment.

# New contributor declaration

- [x] I have written a PR description following the requested rules.
- [ ] I have run `pre-commit run --from-ref upstream/release/3.2.2 --to-ref HEAD`.
- [x] I have added tests.


third_party/ascend/backend/driver.py
```diff
@@ -538,7 +539,7 @@ def _format_of(ty):
             "float": "f",
             "double": "d",
             "long": "l",
-            "int8_t": "b",
```
hongziqi, line 541: This issue also exists in upstream Triton 3.2 NVIDIA/AMD launchers because signed i8 is parsed using the unsigned PyArg_ParseTuple("b") format. Upstream Triton 3.7 resolves it as part of the launcher refactoring by parsing integer arguments directly with PyLong_AsLong(https://github.com/triton-lang/triton/pull/6788/changes).
﻿
To avoid introducing an Ascend-specific parsing path and behavior difference in release/3.2.2, we plan not to backport this fix independently. Instead, the issue will be resolved by synchronizing the upstream 3.7 launcher optimization into the Ascend 3.7 branch, with a signed-i8 regression test added there.

luochao-dev: If release/3.2.2 will continue to be maintained and have new releases, I personally recommend fixing this issue, provided that this part of your code is not shared with the
  community branch and the fix will not introduce side effects such as patch conflicts.

PR #2228, opened 2026-09-17
[ssbuffer](fix) inline transpose yield

lrf22:
# Summary

When a cube-side matmul result is only consumed by a transpose on the vector side (`isAllTranspoesd`), the c→v fixpipe performs the transpose itself (NZ2DN DMA mode) and the consumer-side users of the transposed value are rewired to the transferred tensor. However, when the transposed value is consumed by a yield op (e.g. an `scf.if` / loop yield located outside the consumer block), the generic user-rewriting loop in `insertCubeToVectorTransfer` skipped it — the yield op does not carry the consumer block id, so it never matched `dep.iniConsumerBlockId`. The yield kept referencing the original value and the fixpipe-transposed result was never inlined into the yield.

This PR makes `isAllTransposedInVector` also return the yield op (if any) that uses the transposed value, records it on the C2V `DependencyInfo` as `consumerYieldOp`, and thereby reuses the existing consumer-yield-rewrite mechanism (already in place for iterarg-yield dependencies) so that the yield is rewired to the transferred tensor.

# Impact

- Only affects c→v dependencies where the cube matmul result is exclusively consumed via a vector-side transpose that is in turn used by a yield op; all other c→v paths keep their existing behavior.
- No new fields or interfaces: reuses the existing `DependencyInfo::consumerYieldOp` mechanism.

# New contributor declaration

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests


third_party/ascend/include/DynamicCVPipeline/SplitDataflow/DataDependencyAnalysis.h
```diff
@@ -171,7 +171,7 @@ class DataDependencyAnalysisPass
   bool isValidValueForDependency(mlir::Value value);
   bool isValidScalarDependency(mlir::Value value);
   bool isValid1DValueForDependency(mlir::Value value);
-  bool isAllTransposedInVector(mlir::Value value);
+  std::pair<bool, mlir::Operation *> isAllTransposedInVector(mlir::Value value);
```
olivervnc: 可以遗留：用 std::optional<Operation *>

olivervnc (approved):

third_party/ascend/include/DynamicCVPipeline/SplitDataflow/DataDependencyAnalysis.h
olivervnc: 或者 std::optional<YieldOp>

olivervnc:
补充下 PR 描述

hongziqi (approved):

WuTYSFG (approved):

KanuaK (approved):

PR #2238, opened 2026-09-17
[backend](fix) defer arch-dependent queries out of import time

### LINKED ISSUE #2232: ascend backend cannot be imported when torch_npu is not loaded, which makes triton_key() inconsistent across processes
### Describe the bug

`triton.backends.ascend.runtime` cannot be imported in a process that has not imported `torch_npu`. Any import of that package fails with:

```text
ImportError: libtorch_npu.so: cannot open shared object file: No such file or directory
```

The cause is a **device probe executed at import time** (the Python package itself has no link-time dependency on `libtorch_npu.so`). Installed 3.6.0 nightly, full chain:

```text
site-packages/triton/backends/ascend/runtime/cv_autotune/hardware_consts.py:30
    from ..utils import target                       # this line alone breaks the laziness
site-packages/triton/backends/ascend/runtime/utils.py:71
    __getattr__("target")
site-packages/triton/backends/ascend/runtime/utils.py:36
    _init_npu_params() -> driver.active.get_current_target()
site-packages/triton/backends/ascend/driver.py:267
    get_current_target() -> self.utils.get_arch()
site-packages/triton/backends/ascend/driver.py:185
    get_arch() -> self._load_mod()
site-packages/triton/backends/ascend/driver.py:95
    _load_mod() -> dlopen npu_utils.so
ImportError: libtorch_npu.so: cannot open shared object file: No such file or directory
```

`runtime/utils.py` deliberately exposes `target` / `device` / `prop` / `num_vector_core` / ... as lazy attributes through a module-level `__getattr__`, but `from ..utils import target` asks for the attribute *during import*, so the probe fires immediately. `npu_utils.so` is a JIT-built torch_npu host extension (`NPUUtils._build_or_get_cached_so()` even keys its build cache on `importlib.metadata.version("torch_npu")`), which is why dlopen needs `libtorch_npu.so`.

#### Why this matters beyond "one import fails"

`triton_key()` (`python/triton/runtime/cache.py`) enumerates backend packages with `pkgutil.walk_packages()`, which **imports every package** in order to recurse into it. When that import raises, `walk_packages` skips the whole subtree silently (no `onerror`), and `triton_key()` — itself `functools.lru_cache`d — then returns a key that omits those files. The same installation therefore has two different cache identities:

| process state | `len(triton_key())` | files walked |
|---|---:|---:|
| `TORCH_DEVICE_BACKEND_AUTOLOAD=0`, `torch_npu` not loaded | 3709 | 9 |
| `torch_npu` loaded | 5919 | 43 |

In our environment the missing entries are exactly the 34 modules under `ascend/runtime/` (the blocker is the single package `triton.backends.ascend.runtime`, so its whole subtree is dropped; `runtime/__init__.py` itself is still hashed because `walk_packages` yields the entry before importing it).

This is hit in production by `torch.compile(..., backend="inductor")` on torch_npu: the torch_npu Inductor integration intentionally holds `TORCH_DEVICE_BACKEND_AUTOLOAD=0` while the compile pool starts (to avoid initializing NPU before `fork`), and Inductor calls `triton_key()` in that short-lived "sidecar" process before forking its compile workers (`pre_fork_setup()`), where the value is cached by `lru_cache` and inherited by the forked workers. Consequences seen by users: the autotune best-config cache hits, but the Triton binary cache misses, so kernels are recompiled whenever the compile/load site switches between worker and main process; persistent caches that were supposed to be reusable across processes are split in two.

### Reproduce

```bash
# torch_npu installed but not auto-loaded
TORCH_DEVICE_BACKEND_AUTOLOAD=0 python -c "import triton.backends.ascend.runtime"
```

```python
import hashlib, os
os.environ["TORCH_DEVICE_BACKEND_AUTOLOAD"] = "0"
from triton.runtime.cache import triton_key
k = triton_key()
print(len(k), hashlib.sha256(k.encode()).hexdigest())
```

Per-package import results with `TORCH_DEVICE_BACKEND_AUTOLOAD=0` (one fresh process per package):

| module | result |
|---|---|
| `triton.backends.ascend`, `triton.backends.ascend.compiler`, `triton.backends.ascend.driver` | OK |
| `triton.backends.ascend.runtime` and everything below it (`cv_autotune`, `normalization`, `dsl_analysis`, `hardware_consts`, `tile_generator`, `ubtuner`, `utils`) | FAIL, all with the same `libtorch_npu.so` ImportError |

The submodule failures are inherited from the parent package (importing a submodule imports its parent first).

### Environment details

```text
Triton-Ascend : 3.6.0 (triton_ascend 3.6.0+dev20260909)
Triton        : 3.6.0
torch_npu     : 2.10.0.post5.dev20260901
PyTorch       : 2.10.0+cpu
CANN          : 9.1.0
NPU           : Ascend 950 (Ascend950PR_9579)
OS            : Ubuntu container, Python 3.11
```

### Where the import-time device access is

An AST scan of `third_party/ascend/backend` (module bodies, class bodies, decorators and default arguments — i.e. everything evaluated at import time) finds exactly two sites:

| file | code | blocks `walk_packages`? |
|---|---|---|
| `runtime/cv_autotune/hardware_consts.py:30` and `:62` | `from ..utils import target`; class body `if target.arch.startswith("Ascend910_95") or target.arch.startswith("Ascend950")` | **yes — the only blocker** |
| `runtime/tile_generator.py:36` | `from .utils import (..., num_vector_core, ub_size_in_kbytes, rf_size_in_kbytes)` | no (plain module; `walk_packages` only imports *packages*) |

Isolation confirms it: stubbing out `hardware_consts` makes `triton.backends.ascend.runtime` import successfully, with no other module in the subtree demanding device parameters at import time. `tile_generator` does demand `num_vector_core` at import time when tested on its own, it is just not on the walk's import path.

### Suggested fix

Defer the device probe so that importing the backend never requires the device runtime to be loadable:

```python
# runtime/cv_autotune/hardware_consts.py
# 1) import the module, not the attribute
from .. import utils
# 2) evaluate the arch-dependent defaults when an instance is created
l0c_size: int = field(default_factory=_default_l0c_size)   # same for ub_size / max_l2_size
```

```python
# runtime/tile_generator.py  — move the lazy names into the functions that use them
def _budget():
    from . import utils
    return utils.num_vector_core, utils.ub_size_in_kbytes, utils.rf_size_in_kbytes
```

Both are behaviour-preserving: `HardwareConstraints` still derives its arch-dependent defaults, just at first instantiation instead of at import.

**Verified locally** (in-process prototype, no source change): in a process **without** `torch_npu` loaded, once the import-time probe is neutralised, `triton_key()` becomes `len=5919 / sha256=4820b25022d89652904af8e4df497f0b75aa7cf69151f7945a88375e342562a8` — byte-identical to the key computed in a process **with** `torch_npu` loaded, and the walked file count goes back from 9 to 43. So this change alone removes the key split for the ascend backend.

Independently of the fix above, it would help to make `triton_key()` fail loudly (or enumerate without importing) instead of silently dropping files on `ImportError`; I am happy to send that as a separate PR if you are interested.

### Additional context

- Related: #2200 (same root cause — `walk_packages()` importing `ascend.runtime`; that PR removes a fragile namespace-path `KeyError`, but it does not change which files are hashed).
- After #2200 lands, `cv_autotune/generators` and `cv_autotune/pruning` become regular packages, so their two files will start being enumerated and the key will change once — expected, a one-time rebuild.
- General principle worth stating in the backend API docs: importing a backend package should not require the device runtime to be present, because triton's cache-key computation imports every backend package.

HinPeng:
### What

Defer the device/arch-dependent lookups in the Ascend runtime out of import time, so that `triton.backends.ascend.runtime` can be imported in a process that has not loaded `torch_npu`.

- `runtime/cv_autotune/hardware_consts.py`: the class-body `if target.arch.startswith(...)` block overriding `l0c_size` / `ub_size` / `max_l2_size` forced a target probe during import (the `from ..utils import target` on line 30 is the reported trigger). The overrides move into `field(default_factory=_arch_default)`, so the probe only fires when a caller actually omits the argument.
- `runtime/tile_generator.py`: `num_vector_core`, `ub_size_in_kbytes` and `rf_size_in_kbytes` were bound at module import, which tripped the same lazy-attribute probe. They are now read through the `utils` module at construction time, with `num_vector_core` snapshotted on the `TileGenerator` instance.

Behaviour is unchanged for a given arch: explicit constructor arguments still take precedence over the arch default, and the resolved values are identical.

Fixes https://github.com/triton-lang/triton-ascend/issues/2232


KanuaK (approved):

WuTYSFG (approved):

third_party/ascend/backend/runtime/cv_autotune/hardware_consts.py
```diff
@@ -24,10 +24,18 @@
 including memory limits, cache sizes, and other resource constraints.
 """
 
-from dataclasses import dataclass
```
KanuaK, line 27: supplement testcase later

Issue #2156, opened 2026-09-14
[Bug]: regular @triton.jit helper call inside dynamic tl.range fails Triton-to-Linalg conversion


Liamup777:
### Describe the bug

A regular `@triton.jit` helper that returns a value from two pointer-load branches fails during `ConvertTritonIRToLinalgIR` when it is called inside a dynamic `tl.range` loop.

This was reduced from vLLM's `_thinking_budget_kernel`, where `_load_effective_token` is called while scanning token ranges. All seven tested vLLM input scenarios fail with the same unresolved pointer materialization and invalid `tt.call`.

This appears related to #39, which reports the same conversion failure for `@triton.jit(noinline=True)` functions. The reproducer below does **not** set `noinline=True`; both functions use plain `@triton.jit`. As an ablation, calling the helper directly outside the dynamic `tl.range` succeeds in the same environment.

Minimal reproducer:

```python
import torch
import torch_npu  # noqa: F401
import triton
import triton.language as tl


@triton.jit
def load_effective_token(
    committed_ptr,
    committed_stride,
    pending_ptr,
    request_start,
    request_idx,
    committed_len,
    pos,
):
    if pos < committed_len:
        return tl.load(committed_ptr + request_idx * committed_stride + pos)
    pending_pos = request_start + pos - committed_len + 1
    return tl.load(pending_ptr + pending_pos)


@triton.jit
def reproduce_kernel(
    committed_ptr,
    pending_ptr,
    lengths_ptr,
    request_mapping_ptr,
    output_ptr,
    committed_stride,
):
    token_idx = tl.program_id(0).to(tl.int64)
    request_idx = tl.load(request_mapping_ptr + token_idx)
    committed_len = tl.load(lengths_ptr + request_idx)
    effective_len = committed_len + 1
    total = tl.zeros((), dtype=tl.int32)
    for pos in tl.range(committed_len - 1, effective_len):
        token = load_effective_token(
            committed_ptr,
            committed_stride,
            pending_ptr,
            token_idx,
            request_idx,
            committed_len,
            pos,
        )
        total += token
    tl.store(output_ptr + token_idx, total)


def main():
    device = "npu:1"
    committed = torch.tensor([[10, 11]], dtype=torch.int32, device=device)
    pending = torch.tensor([11, 20], dtype=torch.int32, device=device)
    lengths = torch.tensor([2], dtype=torch.int32, device=device)
[...]

Issue #39, opened 2026-04-27
Supporting  `noinline` kernel functions


dgarus:
### Describe the bug

When I try to use `@triton.jit(noinline=True)`, I get the following error:
```
python3 ./scripts/reproducer.py
loc("/workspace/./scripts/reproducer.py":14:0): error: failed to legalize unresolved materialization from ('memref<?xf32>') to '!tt.ptr<f32>' that remained live after conversion
loc("/workspace/./scripts/reproducer.py":14:0): error: failed to apply Conversion Patterns
loc("/workspace/./scripts/reproducer.py":18:23): error: 'tt.call' op 'level_1__Pfp32_i64__' does not reference a valid function
Traceback (most recent call last):
  File "/usr/local/python3.11.14/lib/python3.11/site-packages/triton/compiler/compiler.py", line 289, in compile
    next_module = compile_ir(module, metadata)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/python3.11.14/lib/python3.11/site-packages/triton/backends/ascend/compiler.py", line 1058, in <lambda>
    stages["ttadapter"] = lambda src, metadata: ttir_to_linalg(
                                                ^^^^^^^^^^^^^^^
  File "/usr/local/python3.11.14/lib/python3.11/site-packages/triton/backends/ascend/compiler.py", line 162, in ttir_to_linalg
    pm.run(mod)
RuntimeError: PassManager::run failed

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/workspace/./scripts/reproducer.py", line 36, in <module>
    main()
  File "/workspace/./scripts/reproducer.py", line 27, in main
    debug_kernel[(1,)](x, out)
  File "/usr/local/python3.11.14/lib/python3.11/site-packages/triton/runtime/jit.py", line 333, in <lambda>
    return lambda *args, **kwargs: self.run(grid=grid, warmup=False, *args, **kwargs)
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/python3.11.14/lib/python3.11/site-packages/triton/runtime/jit.py", line 622, in run
    kernel = self._do_compile(key, signature, device, backend, target, constants, options, configs[0], warmup)
[...]

PR #2312, opened 2026-09-20
[TritonToLinalg](feat) Support multi-return kernel call  in kernel

TecJesh:
### Related Issue
https://github.com/triton-lang/triton-ascend/issues/2314
https://github.com/triton-lang/triton-ascend/issues/2156

### Summary

Lower tt.call ops that survive TTIR inlining into real device-side func.call in TritonToLinalg, fixing compile failures when a kernel calls a @triton.jit helper with multiple returns (branches) from inside a kernel.

1. Add TritonCallOpPattern: tt.call → func.call, forwarding the caller's trailing 6 grid args (program_id×3 + num_programs×3)
2. Stamp global_kernel only on the public entry kernel; skip declarations in addProgramInfo.
3. Preserve function visibility in convertTTFunc; inject host-side workspace/sync-block-lock args only into the public kernel; skip declarations in the step-8 walk.

New tests in test_kernel_function_call.py cover top-level, for-loop, nested-loop and nested-if calls.

### Background

- In make_ttir, the generic inliner cannot inline a multi-block callee (branches / multiple returns) into a single-block region such as a scf.for body or scf.if region, so calls inside loops / nested branches are retained as call nodes in TTIR. Verified that top-level calls to such
  helpers are inlined, so they never reach downstream passes.
- Verified that the GPU backend produces identical TTIR — the loop-body tt.call is retained there as well. The divergence is purely backend-side: NVIDIA's TritonToTritonGPU/TritonGPUToLLVM has handling for tt.call (lowers it to an LLVM device function call), while Ascend's TritonToLinalg
  had no handling for function calls in loop bodies → dialect conversion fails (memref materialization crash at the call boundary

### Why

- Lowering to func.call instead of inlining: honoring the call semantics is the only correct option once inlining is illegal.
- Forward the caller's grid args: Ascend has no device-side grid intrinsic — tl.program_id() lowers to a read of the function's tse values only exist at the entry kernel (host-filled at launch). Forwarding is what inlining would have done implicitly.
- global_kernel only on the public kernel: bishengir's AdaptTritonKernel treats every function with this attribute as a device entry and rewrites its argument list, which would corrupt a private callee's signature called via func.call.
- Preserve visibility + host args only on the kernel: helpers must stay private so host-side workspace/sync-block-lock args are iernel.
- Skip declarations: addProgramInfo's getBody().front().addArgument asserts on an empty body.

# New contributor declaration

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests


third_party/ascend/unittest/pytest_ut/test_kernel_function_call.py
```diff
@@ -0,0 +1,232 @@
...
+            request_idx,
+            committed_len,
+            pos,
+        )
+        total += token
+    tl.store(output_ptr + token_idx, total)
+
+
+@triton.jit
+def helper_top(x_ptr, idx):
```
hongziqi, line 74: Suggest adding some MLIR test cases

third_party/ascend/unittest/pytest_ut/test_kernel_function_call.py
```diff
@@ -0,0 +1,232 @@
...
+import pytest
+import torch
+import torch_npu  # noqa: F401
+import triton
+import triton.language as tl
+import test_common
+
+
+@triton.jit
+def load_effective_token(
```
hongziqi, line 30: The current cases only have multiple returns for scalar, which should cover other ranges such as tensor, and it is necessary to test whether the cases with noinline=true are working properly now

third_party/ascend/unittest/pytest_ut/test_kernel_function_call.py
```diff
@@ -0,0 +1,232 @@
...
+    committed_ptr,
```
hongziqi, line 31: Suggest linking issue: https://github.com/triton-lang/triton-ascend/issues/39

